PR #10, opened 2023-12-15
Allow environment variables in compose file

haarts:


nanderstabel (approved):
Useful addition of ensuring that the `AGENT_APPLICATION_HOST` will be set.

Thank you for this contribution!

PR #15, opened 2024-01-12
feat: support application base path, add Docker build scripts

Norlock:
# Description of change
This includes all the code that is needed to deploy on Kubernetes. It also has a relative path included for if you host it with a relative path. 

There is another PR coming from Nander to fix the urls that are returned from the server that fixes the urls that will be returned from the agent.

I made sure it got deployed on Kubernetes automatically, and also tested the local settings without relative path.


agent_application/src/main.rs
```diff
@@ -10,7 +10,7 @@ use agent_store::{in_memory, postgres};
 use lazy_static::lazy_static;
 
 lazy_static! {
-    static ref HOST: url::Url = format!("http://{}:3033/", config!("host").unwrap()).parse().unwrap();
+    static ref HOST: url::Url = format!("https://{}:3033/", config!("host").unwrap()).parse().unwrap();
```
daniel-mader: Will this still work with a local dev environment (non-https)?

agent_application/docker/docker-compose.yml
```diff
@@ -34,5 +34,7 @@ services:
     environment:
       AGENT_CONFIG_LOG_FORMAT: json
       AGENT_CONFIG_EVENT_STORE: postgres
+      # Relative paths can't start or end with '/'
+      AGENT_RELATIVE_PATH: "unicore"
```
daniel-mader: A common naming for this is `BASE_PATH`. Do you wanna update to `AGENT_APPLICATION_BASE_PATH`?

agent_application/docker/README.md
```diff
@@ -22,4 +22,13 @@ Inside the folder `/agent_application/docker`:
 docker compose up -d
 ```
 
+
 3. The REST API will be served at `http://0.0.0.0:3033`
+
+---
+**NOTE**
+
+When you set the AGENT_RELATIVE_PATH to for example: "unicore"
```
daniel-mader: Also update to `AGENT_APPLICATION_BASE_PATH`?

daniel-mader (changes requested):

agent_api_rest/src/lib.rs
```diff
@@ -16,21 +16,44 @@ use credential_issuer::{
...
+    let base_path: Option<String> = env::var_os("AGENT_APPLICATION_BASE_PATH").map(|os| {
+        let err_msg = "can't parse AGENT_APPLICATION_BASE_PATH";
+        let base_path = os.into_string().expect(err_msg);
+
+        if base_path.starts_with("/") || base_path.ends_with("/") {
+            println!("AGENT_APPLICATION_BASE_PATH can't start or end with '/'");
+        }
+
+        base_path
+    });
```
nanderstabel: To be consistent elsewhere in this repo, we should not make use of system environment variables, but rather only use environment variables that are listed in the `.env` file (or `agent_shared/tests/.env.test` file when running tests). 

So I would suggest using something like this here instead using our `config!` macro:
```rust
    let base_path: Option<String> = config!("base_path")
        .map(|base_path| {
            if base_path.starts_with("/") || base_path.ends_with("/") {
                warn!("AGENT_APPLICATION_BASE_PATH can't start or end with '/'");
            }

            base_path
        })
        .ok();
```

`config!("base_path")` appends `"base_path"` to the package name in which it is called (`"application_agent"`) in order to look for `AGENT_APPLICATION_BASE_PATH` in the `env` file (or `.env.test` file.

agent_application/src/main.rs
```diff
@@ -20,15 +17,31 @@ async fn main() {
...
+        }
+    } else {
+        tracing_subscriber::fmt::init()
     }
 
-    initialize(state.clone(), startup_commands(HOST.clone())).await;
+    let url = env::var_os("AGENT_APPLICATION_URL")
+        .expect("AGENT_APPLICATION_URL is not set")
+        .into_string()
+        .unwrap();
```
nanderstabel:
```suggestion
let url = config!("url").expect("AGENT_APPLICATION_URL is not set");
```

agent_shared/src/url_utils.rs
```diff
@@ -0,0 +1,82 @@
+pub trait AddFunctions {
```
nanderstabel: Perhaps this is a bit too generic of a name for this trait? I think using `UrlUtils` or something like that is a bit more explicit.

agent_application/src/main.rs
```diff
@@ -20,15 +17,31 @@ async fn main() {
...
+    let url = url::Url::parse(&url).unwrap();
+
+    initialize(state.clone(), startup_commands(url)).await;
+
+    let server = "0.0.0.0:3033".parse().unwrap();
 
-    axum::Server::bind(&"0.0.0.0:3033".parse().unwrap())
-        .serve(app(state).into_make_service())
+    axum::Server::bind(&server)
+        .serve(app(state.clone()).into_make_service())
```
nanderstabel: No need to clone `state` here

nanderstabel (changes requested):
Looks good, just left some remarks.

Also `cargo clippy` is still giving some minor warnings, i.e:
```
 --> agent_shared/src/url_utils.rs:33:29
   |
33 |         if path.starts_with("/") {
   |                             ^^^ help: try using a `char` instead: `'/'`
   |
```

agent_shared/src/url_utils.rs
daniel-mader, line 1: Have you explored other alternatives such as the official [path_segments_mut()](https://docs.rs/url/latest/url/struct.Url.html#method.path_segments_mut) before introducing our own custom functions?

agent_application/src/main.rs
```diff
@@ -20,15 +17,31 @@ async fn main() {
         _ => in_memory::ApplicationState::new(vec![Box::new(SimpleLoggingQuery {})], IssuanceServices {}).await,
     };
 
-    match config!("log_format").unwrap().as_str() {
-        "json" => tracing_subscriber::fmt().json().init(),
-        _ => tracing_subscriber::fmt::init(),
+    if let Ok(val) = config!("log_format") {
```
daniel-mader: Rename `val` to `log_format`?

agent_application/docker/docker-compose.yml
```diff
@@ -27,12 +27,17 @@ services:
...
+      context: ../..
+      dockerfile: ./agent_application/docker/Dockerfile
+    #restart: always
     ports:
       - 3033:3033
     environment:
       AGENT_CONFIG_LOG_FORMAT: json
       AGENT_CONFIG_EVENT_STORE: postgres
-      AGENT_APPLICATION_HOST: ${AGENT_APPLICATION_HOST:?set it please}
+      # Relative paths can't start or end with '/'
```
daniel-mader: Is this comment still applicable?

agent_application/docker/README.md
```diff
@@ -12,14 +12,23 @@ docker build -f docker/Dockerfile -t ssi-agent ..
...
-1. _Inside `docker-compose.yml` replace the value `<your-local-ip>` for the environment variable `AGENT_APPLICATION_HOST` with your actual local ip address (such as 192.168.1.234)_
+1. _Inside `docker-compose.yml` replace the environment value: `AGENT_APPLICATION_URL` with your actual local ip address (such as http://192.168.1.234:3033)
 2. Optionally, add the following environment variables:
     - `AGENT_ISSUANCE_CREDENTIAL_NAME`: To set the name of the credentials that will be issued.
     - `AGENT_ISSUANCE_CREDENTIAL_LOGO_URL`: To set the URL of the logo that will be used in the credentials.
 3. To start the **SSI Agent**, a **Postgres** database along with **pgadmin** (Postgres Admin Interface) simply run:
 
 ```bash
-docker compose up -d
+docker compose up
```
daniel-mader, line 22: Why remove the `-d`?

Changelog.md
daniel-mader, line 1: I think it's a convention to use all caps for that filename like so: `CHANGELOG.md`

daniel-mader (changes requested):

agent_application/docker/README.md
Norlock: Because when developing i want to restart whenever I make changes, and I want to see the logs. Why would you want to run it in the background? then you can't quit easily

agent_shared/src/url_utils.rs
Norlock: Yes but its not very convenient: https://docs.rs/url/latest/url/struct.PathSegmentsMut.html.

agent_application/docker/docker-compose.yml
Norlock: I will make a fix that it will automatically remove.

agent_shared/src/url_utils.rs
nanderstabel: `path_segments_mut()` is pretty inconvenient I agree.

README.md
```diff
@@ -8,6 +8,10 @@
 
 Build and run the **SSI Agent** in a local Docker environment following [these steps](./agent_application/docker/README.md).
 
+## Changelog
+
+From time to time breaking changes can occur. Please make sure you read [Changelog.md](./Changelog.md) before updating.
```
daniel-mader: Please update the links to the updated filename.

.pipeline/.gitignore
```diff
@@ -0,0 +1,2 @@
+kustomize*
```
daniel-mader: Is the ignore for `kustomize*` still required?

agent_application/docker/README.md
```diff
@@ -12,14 +12,23 @@ docker build -f docker/Dockerfile -t ssi-agent ..
 
 Inside the folder `/agent_application/docker`:
 
-1. _Inside `docker-compose.yml` replace the value `<your-local-ip>` for the environment variable `AGENT_APPLICATION_HOST` with your actual local ip address (such as 192.168.1.234)_
+1. _Inside `docker-compose.yml` replace the environment value: `AGENT_APPLICATION_URL` with your actual local ip address (such as http://192.168.1.234:3033)
```
daniel-mader: Remove `_` at the beginning or add another `_` at the end please.

agent_application/docker/README.md
```diff
@@ -12,14 +12,23 @@ docker build -f docker/Dockerfile -t ssi-agent ..
...
     - `AGENT_ISSUANCE_CREDENTIAL_NAME`: To set the name of the credentials that will be issued.
     - `AGENT_ISSUANCE_CREDENTIAL_LOGO_URL`: To set the URL of the logo that will be used in the credentials.
 3. To start the **SSI Agent**, a **Postgres** database along with **pgadmin** (Postgres Admin Interface) simply run:
 
 ```bash
-docker compose up -d
+docker compose up
 ```
 
+
```
daniel-mader, line 26: Why another empty line here? Also, I think the enumeration is incorrect, there's `3.` twice.

agent_application/docker/docker-compose.yml
```diff
@@ -27,12 +27,16 @@ services:
     depends_on:
       - cqrs-postgres-db
   ssi-agent:
-    image: ssi-agent
-    restart: always
+    #image: ssi-agent
+    build:
+      context: ../..
+      dockerfile: ./agent_application/docker/Dockerfile
+    #restart: always
```
daniel-mader: Can we switch to something such as `restart: "${DOCKER_RESTART_POLICY:always}"`, so we don't have to comment it out for dev/prod environments?

agent_application/docker/docker-compose.yml
```diff
@@ -27,12 +27,16 @@ services:
     depends_on:
       - cqrs-postgres-db
   ssi-agent:
-    image: ssi-agent
-    restart: always
+    #image: ssi-agent
```
daniel-mader, line 30: Maybe we should rather introduce a separate `docker-compose.dev.yml` or something. We cannot expect someone who just wants to run the application to know where and what to uncomment.

agent_shared/src/url_utils.rs
```diff
@@ -0,0 +1,120 @@
...
+             let res = url::Url::parse(&format!("{}/", url)).unwrap();
+             tracing::info!("res: {:?}", res);
+             res
+        } else {
+            url.clone()
+        }
+}
+
+impl UrlAddFunctions for url::Url {
+    fn add_path(&self, path: &str) -> url::Url {
```
daniel-mader: Can we rename to `append_path_segment()`?

agent_shared/src/url_utils.rs
```diff
@@ -0,0 +1,120 @@
...
+            Ok(url) => url,
+            Err(err) => {
+                let err_str = format!("Path can't be added: {:?}\n{:?}", path, err);
+                tracing::error!("{:?}", &err_str);
+                panic!("{:?}", &err_str);
+            }
+        }
+    }
+
+    fn add_file(&self, file: &str) -> url::Url {
```
daniel-mader: I think this function name is a little bit misleading, `add_file()` sounds to me like I would actually supply some some bytes to create an actual file. Maybe we could rename to `append_file_name()` or something alike?

agent_shared/src/url_utils.rs
```diff
@@ -0,0 +1,120 @@
+pub trait UrlAddFunctions {
```
daniel-mader: Rename to `UrlAppendHelpers`?

agent_api_rest/src/lib.rs
```diff
@@ -21,31 +22,76 @@ use offers::offers;
 pub const AGGREGATE_ID: &str = "agg-id-F39A0C";
 
 pub fn app(state: ApplicationState<IssuanceData, IssuanceDataView>) -> Router {
+    let base_path = config!("base_path").map(|mut base_path| {
```
daniel-mader: Can this be extracted in its own helper function? I feel like it clutters the `app()` function.

agent_application/docker/docker-compose.yml
Norlock: No we shouldn't in my opinion. If people want to have a production environment they have to setup themselves. Not just use ours. Thats also why I don't think we should restart, its just for us. Or they can use it as a template but nothing more than that.

agent_application/docker/docker-compose.yml
Norlock: I added it, not sure if I like it. I want it to crash if something goes wrong.

daniel-mader: Okay, fair. I'm just not the biggest fan of having to comment in/out in different files (and always have to remember to set it back before I make a commit), but rather set it through some env variable that I can define for myself and that is untracked.

agent_application/docker/docker-compose.yml
daniel-mader: Okay, we could remove `#image: ssi-agent` then altogether, right?

.pipeline/test_build.sh
daniel-mader, line 1: I think the env variables should be set through an untracked `.env` file instead.

daniel-mader (approved):

nanderstabel (approved):

PR #26, opened 2024-02-13
refactor: add `Offer`, `Credential` and `ServerConfig` aggregates

### LINKED ISSUE #25: Split `IssuanceData` aggregate into multiple aggregates
### Description

Currently we make use of one single `IssuanceData` aggregate which consists of the complete `agent_issuance` data. The result of this is that all this data is persisted as one single blob. We need to split this aggregate into multiple aggregates according to Domain Driven Design principles. 

Aggregates we've identified so far are (may change):
- `Credential`
- `Offer`
- `ServerConfig`

### Motivation

Splitting the `IssuanceData` aggregate will result in cleaner separation of data. Getting rid of one single blob of data will also increase performance of the agent.

### Resources

_No response_

### To-do List

- [ ] Split `IssuanceData` aggregate into multiple aggregates

nanderstabel:
# Description of change
This change splits the current `IssuanceData` aggregate into 3 aggregates:
- `Credential`
- `Offer`
- `ServerConfig`

These aggregates are implemented according to [DDD](https://en.wikipedia.org/wiki/Domain-driven_design) principles.

In order to do this, as well as allowing for `PreAuthorizedCode` and `AccessToken` to be queried, the `ApplicationState` struct has been altered to represent the clear separation of responsibility of Commands and Queries (according to the [CQRS](https://martinfowler.com/bliki/CQRS.html) pattern):
```rust
#[derive(Clone)]
pub struct ApplicationState {
    pub command: CommandHandlers,
    pub query: Queries,
}
```

fixes #25 

- All the aggregates commands are tested in `agent_issuance` in a similar fashion in to they were already tested before (with some alterations)
- All endpoints are tested in `agent_api_rest`
- The postman collection is slightly updated: `preAuthorizedCode` is removed and a test script that fetches the {{preAuthorizedCode}} variable from the `offers` endpoint response is added.


agent_api_rest/src/credential_issuer/credential.rs
```diff
@@ -12,62 +15,102 @@ use axum::{
...
+        }
     };
 
-    match command_handler(AGGREGATE_ID.to_string(), &state, command).await {
-        Ok(_) => StatusCode::NO_CONTENT.into_response(),
-        Err(err) => {
-            println!("Error: {:#?}\n", err);
-            (StatusCode::BAD_REQUEST, err.to_string()).into_response()
+    // Use the `offer_id` to get the `credential_ids` from the `OfferView`.
+    let credential_ids = match query_handler(&offer_id, &state.query.offer).await {
```
tensor-programming, line 37: These match statements feel a bit weird, especially given the TODO above.  Its strange that these functions return this `IntoResponse` trait type instead of a concrete implementation because you could also explicitly handle the errors.  The code obviously works and it is idiomatic but the added complexity could cause trouble in the future.

agent_api_rest/src/credential_issuer/credential.rs
```diff
@@ -12,62 +15,102 @@ use axum::{
...
+        Ok(Some(OfferView { credential_ids, .. })) => credential_ids,
+        _ => {
+            info!("Returning 500");
```
tensor-programming: You could create a simple macro to combine the info call with the error return type so that you don't have to explicitly call them every time. `info!("Returning 400")` appears 4 times while `info!("Returning 500")` appears a total of 15 times with every single one returning the `StatusCode` into response call. 

With the macro you could probably simplify these match statements as well.

agent_api_rest/src/credentials.rs
```diff
@@ -1,99 +1,170 @@
...
+    } else {
+        info!("Returning 400");
+        return (StatusCode::BAD_REQUEST, "credential is required".to_string()).into_response();
     };
 
-    match command_handler_without_id(&state, command).await {
+    let credential_id = uuid::Uuid::new_v4().to_string();
+
+    // Create an unsigned credential.
+    match command_handler(
```
tensor-programming: For the `command_handler` calls, you have these empty match patterns.  Since you only care about whether or not they return an Err, you could simplify them all by using `if let Err(_)` or one of the result helper functions.  It would make the code slightly clearer though, again, the issue comes back to the fact that the outer function can't return a result type.

agent_issuance/src/handlers.rs
```diff
@@ -1,14 +1,22 @@
...
-use tracing::{debug, error};
+use tracing::{debug, error, info};
 
-pub async fn query_handler<A: Aggregate, V: View<A>>(
-    credential_id: String,
-    state: &ApplicationState<A, V>,
-) -> Result<Option<V>, PersistenceError> {
-    match state.load(&credential_id).await {
+/// The `query_handler` function is used to query the view repository for a specific view.
+pub async fn query_handler<A, V>(
```
tensor-programming, line 11: Is there really value in having the option here?  It seems like every time you call this function, you are only looking to see if the return is `Ok(Some(value))` which means `Ok(None)` and `Err(...)` end up resulting in the same kind of error state.  I understand wanting the granularity that this gives you, but the Option could be converted into an error or dealt with higher up in the call stack to simplify things.

tensor-programming (commented):
Having gone over the code, it looks relatively clean and easy to follow.  The `cqrs` pattern makes sense for this particular case but its implementation could be slightly better for the sake of the error and logging story.  Outside of that, everything looks good; there is no credential leaking or anything of that nature.

agent_api_rest/src/credential_issuer/credential.rs
daniel-mader: Couldn't you just build a custom `TraceLayer` as well and hook directly into the web framework? [See this example from axum.](https://github.com/tokio-rs/axum/blob/main/examples/tracing-aka-logging/src/main.rs) I feel like proper tracing adjusts to different envs more easily and can also push to metrics collectors more easily in the future.

tensor-programming: Yeah, well its probably a good idea to include real tracing rather than these explicit info calls.

agent_issuance/src/handlers.rs
nanderstabel: Indeed I didn't really make use of `Ok(None)` previously but now with some updates I made it makes a bit more sense I think, e.g.:
```rust
#[axum_macros::debug_handler]
pub(crate) async fn get_credentials(
    State(state): State<ApplicationState>,
    Path(credential_id): Path<String>,
) -> Response {
    // Get the credential if it exists.
    match query_handler(&credential_id, &state.query.credential).await {
        Ok(Some(CredentialView { data: Data { raw }, .. })) => (StatusCode::OK, Json(raw)).into_response(),
        Ok(None) => StatusCode::NOT_FOUND.into_response(),
        _ => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
    }
}
```

This way the `command_handler` aids us in making the distinction between server errors and client errors.

agent_api_rest/src/credentials.rs
nanderstabel: Good point, I've now changed the match patterns to `if command_handler(...).await.is_err() { .. }`

agent_api_rest/src/credential_issuer/credential.rs
nanderstabel: Yeah I ended up using a `TraceLayer` which is indeed a much cleaner solution :)

agent_api_rest/src/credential_issuer/credential.rs
nanderstabel: this match statement currently looks like this:
```rust
    let credential_ids = match query_handler(&offer_id, &state.query.offer).await {
        Ok(Some(OfferView { credential_ids, .. })) => credential_ids,
        _ => return StatusCode::INTERNAL_SERVER_ERROR.into_response(),
    };
```

where eventually all (of not most) of the `INTERNAL_SERVER_ERROR` responses must be replaced with the official [oid4vci](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-credential-error-response) errors. In our current `oid4vci` crate we mainly focused on happy paths, but once the error response are defined in that crate, I will apply them here (not sure yet how). For the time being I think returning 'blanco' ` StatusCode::INTERNAL_SERVER_ERROR` responses this way will do.

daniel-mader (approved):
Awesome! 🥳

PR #71, opened 2024-06-05
feat: use polling strategy for JIT credentials

### LINKED ISSUE #70: Use polling strategy instead of `EXTERNAL_SERVER_RESPONSE_TIMEOUT_MS` for JIT credentials
### Description

Currently our `/v1/credentials` endpoint always sleeps for 250 ms to accommodate for JIT credentials:

```rust
    // This ensures that the server waits for a sufficient duration to potentially receive a credential from an external
    // server.
    std::thread::sleep(std::time::Duration::from_millis(EXTERNAL_SERVER_RESPONSE_TIMEOUT_MS));
```

We need to use a database polling strategy instead so there won't be any needles delay in the endpoint.

### Motivation

This will result in a smoother credential issuance experience.

### Resources

n/a

### To-do List

- [ ] Use polling strategy

nanderstabel:
# Description of change
In the JIT Credential scenario when the wallet sends its Credential Request to UniCore's credential endpoint, the Request will be verified and the `CredentialRequestVerified` event will be published to an external system. 

Then, the endpoint needs to wait for the JIT Credential to arrive. With the current change the endpoint will poll the database every 50 ms to check if the credential arrived. 

By default:
```rust
const DEFAULT_EXTERNAL_SERVER_RESPONSE_TIMEOUT_MS: u128 = 1000;
```
Can be overwritten by setting:
```env
AGENT_API_REST_EXTERNAL_SERVER_RESPONSE_TIMEOUT_MS=500
```

in the env file/docker-compose file.

Note: while implemented I ran into a bug, which is that when events are published by the `EventPublisherHttp`, reqwest::Client` would await for a response and therefore blocking the flow.  I fixed that by sending the request in a separate thread like:
```rust
                // Send the request in a separate thread so that we don't have to await the response in the current thread.
                tokio::task::spawn(async move {
                    request.send().await.ok();
                });
```

The side-effect of this is that several tests related to `EventPublisherHttp` had to be updated to be able to handle multiple threads: `#[tokio::test(flavor = "multi_thread")]`

fixes #70 
future improvement: #75 

Added the `case::should_panic_due_to_timout` to `test_credential_endpoint`.
Also manually tested utilizing UniMe and https://github.com/nanderstabel/test-server


agent_api_rest/src/issuance/credential_issuer/credential.rs
```diff
@@ -16,9 +21,11 @@ use axum::{
 use axum_auth::AuthBearer;
 use oid4vci::credential_request::CredentialRequest;
 use serde_json::json;
-use tracing::info;
+use tokio::time::sleep;
+use tracing::{error, info};
 
-const EXTERNAL_SERVER_RESPONSE_TIMEOUT_MS: u64 = 250;
+const DEFAULT_EXTERNAL_SERVER_RESPONSE_TIMEOUT_MS: u128 = 1000;
+const POLLING_INTERVAL_MS: u64 = 50;
```
daniel-mader: Isn't that a little too often? That's 20 polls per second ...

agent_api_rest/src/verification/relying_party/redirect.rs
```diff
@@ -187,6 +187,9 @@ pub mod tests {
         request(&mut app, state.clone()).await;
         redirect(&mut app, state).await;
 
+        // Wait for the request to arrive at the mock server endpoint.
+        std::thread::sleep(std::time::Duration::from_millis(10));
```
daniel-mader: Is `10ms` safe for that test under all circumstances? Why not increase it to `100ms` to be even more certain?

agent_event_publisher_http/src/lib.rs
```diff
@@ -187,6 +187,9 @@ mod tests {
         // Dispatch the event.
         publisher.offer.as_ref().unwrap().dispatch("view_id", &events).await;
 
+        // Wait for the request to arrive at the mock server endpoint.
+        std::thread::sleep(std::time::Duration::from_millis(10));
```
daniel-mader: See other comment

agent_api_rest/src/issuance/credential_issuer/credential.rs
```diff
@@ -242,13 +276,19 @@ mod tests {
...
+    #[case::with_external_server(true, false, 0)]
+    #[case::with_external_server_and_self_signed_credential(true, true, 0)]
+    #[should_panic(expected = "assertion `left == right` failed\n  left: 500\n right: 200")]
+    #[case::should_panic_due_to_timout(true, false, DEFAULT_EXTERNAL_SERVER_RESPONSE_TIMEOUT_MS + 100)]
     #[serial_test::serial]
-    #[tokio::test]
-    #[tracing_test::traced_test]
-    async fn test_credential_endpoint(#[case] with_external_server: bool, #[case] is_self_signed: bool) {
+    #[tokio::test(flavor = "multi_thread")]
+    // #[tracing_test::traced_test]
```
daniel-mader: Can this be removed?

agent_api_rest/src/issuance/credential_issuer/credential.rs
```diff
@@ -342,7 +382,7 @@ mod tests {
                     .unwrap(),
             )
             .await
-            .unwrap();
+            .expect("hello");
```
daniel-mader: Is this intentional?

daniel-mader (approved):

PR #85, opened 2024-06-20
feat: add support for multiple credential formats

### LINKED ISSUE #67: Support multiple credential formats
### Description

UniCore needs to be able to support multiple credential formats (OpenBadges, EDC, W3C VC, etc...)

### Motivation

Currently UniCore only supports OpenBadgesV3, partially hardcoded. we need to make UniCore more Credential format agnostic in order to be able to support many different formats.

### Resources

n/a

### To-do List

- [ ] Implement a credential format agnostic solution

nanderstabel:
# Description of change
Adds support for multiple credential formats.

fixes #67 

- Manually tested using UniMe
- Added several new unit tests
- Improved the Postman Collection


agent_api_rest/src/issuance/credentials.rs
```diff
@@ -58,17 +62,38 @@ pub(crate) async fn credentials(
 
     let credential_id = uuid::Uuid::new_v4().to_string();
 
+    let credential_configuration = match query_handler(SERVER_CONFIG_ID, &state.query.server_config).await {
+        Ok(Some(ServerConfigView {
+            credential_issuer_metadata:
+                Some(CredentialIssuerMetadata {
+                    credential_configurations_supported,
```
Norlock, line 69: maybe its style difference but I find the style a bit hard to read with all the nesting. Don't you think if we make this a bit more compact, less indented it would be easier to read?

agent_issuance/src/credential/queries.rs
```diff
@@ -15,11 +16,10 @@ impl View<Credential> for CredentialView {
         match &event.payload {
             CredentialEvent::UnsignedCredentialCreated {
                 data,
-                credential_format_template,
+                credential_configuration,
             } => {
                 self.data.replace(data.clone());
-                self.credential_format_template
-                    .replace(credential_format_template.clone());
+                self.credential_configuration = credential_configuration.clone();
```
Norlock, line 22: Is clone needed here?

agent_issuance/Cargo.toml
```diff
@@ -11,8 +11,12 @@ agent_secret_manager = { path = "../agent_secret_manager" }
 async-trait.workspace = true
 axum.workspace = true
 cqrs-es.workspace = true
+chrono = "0.4"
+types-ob-v3 = { git = "https://github.com/impierce/digital-credential-data-models.git" }
```
Norlock: Add =>   , branch = "dev". So we know its not the main branch ;).

agent_verification/src/authorization_request/aggregate.rs
```diff
@@ -381,13 +381,13 @@ pub mod tests {
...
                 .to_string();
         static ref SIGNED_AUTHORIZATION_REQUEST_OBJECT_DID_IOTA: String =
```
Norlock: I know this is not changed but maybe we can change this into const :&str. I think its more applicable here

agent_issuance/src/credential/queries.rs
nanderstabel: `credential_configuration` is a reference here so yes a clone is needed.

agent_issuance/Cargo.toml
nanderstabel: Hmm since we stil have not put a versioning logic in place I think we're better of using the exact `rev` :ok_hand:

agent_verification/src/authorization_request/aggregate.rs
nanderstabel: You're right, changed it :ok_hand:

agent_api_rest/src/issuance/credentials.rs
nanderstabel: hmm yeah I find it quite easy too read and I wouldn't now how to make this more compact other than implementing some method on `ServerConfigView` (which I would not prefer) :thinking: 

Do you have any suggestions? :)

Norlock: Yes lets wait when we have a result type to automatically return from functions, I think that will be the biggest part

Norlock (dismissed):

Norlock (approved):
LGTM

PR #105, opened 2024-08-07
perf: cache DID documents (optional)

daniel-mader:
# Description of change
Enables an optional in-memory cache for DID documents (implemented in `did-manager`). It enhances the speed of interactions for DIDs that need to be resolved from some trusted network.

**Enable this feature with caution! Caching DID documents can cause unexpected behavior.**

If you decide to enable the cache, you should be aware of the implications and keep the TTL to a minimum.

n/a

Enable the cache with different parameters and exploratively test its behavior.


agent_application/Cargo.toml
```diff
@@ -16,6 +16,7 @@ agent_verification = { path = "../agent_verification" }
 axum.workspace = true
 did_manager.workspace = true
 identity_document = { version = "1.3" }
+identity_verification = { version = "1.3" }
```
nanderstabel: This in `did-manager` introduced an openssl dependency in UniMe before. Although there doesn't seem to be any issue here, maybe it is smart to pre-emptively set this to `identity_verification = { version = "1.3", default-features = false }` so we don't run in the same issue again once we add UniCore to UniMe.

(perhaps also smart to add it to the workplace.dependencies since `agent_shared` also depends on it)

agent_shared/Cargo.toml
```diff
@@ -16,6 +16,7 @@ identity_credential.workspace = true
 identity_did = { version = "1.3" }
 identity_document = { version = "1.3" }
 identity_storage = { version = "1.3" }
+identity_verification = { version = "1.3" }
```
nanderstabel: See above comment

nanderstabel (approved):

Issue #40, opened 2024-04-10
Problem Details for HTTP APIs (RFC 7807)


daniel-mader:
### Description

Readability and conciseness of errors via the HTTP API (`agent_api_rest`) can be improved when following the [Problem Details for HTTP APIs (RFC 7807)](https://datatracker.ietf.org/doc/html/rfc7807). It introduces a "standard" way to report errors back to the caller.

### Motivation

Enhanced API experience for the consumer.
We also want a more easy way to handle errors other than manually returning

```
return StatusCode::INTERNAL_SERVER_ERROR.into_response();
```

There are multiple options to explore:

- make use of the framework's error handling
- propagate the error (using `thiserror` and `?`) and build a custom error handler / mapper

### Requirements

1. read and understand the scope of the RFC 7807
2. search for a crate to make implementation easier (possibly [this one](https://crates.io/crates/http-api-problem))
3. implement and test

### Are you planning to contribute this in a PR?

Yes

nanderstabel:
(partially) depends on:https://github.com/impierce/openid4vc/issues/78

M-Adam-Hus:
![](https://github.trello.services/images/mini-trello-icon.png) [Problem Details for HTTP APIs (RFC 7807)](https://trello.com/c/rDTtS86f/8-problem-details-for-http-apis-rfc-7807)

PR #113, opened 2024-08-26
feat: introduce holder capabilities

### LINKED ISSUE #109: Offer credentials between UniCore instances
### Description

A UniCore instance should be able to offer credentials not only to a wallet (UniMe), but also to another UniCore instance.

The receiving UniCore should have an endpoint open waiting for OID4VCI offers to be posted. It will not immediately accept the offer, but wait for one of it's admin users to accept or decline it.

> ~_All API routes should be prefixed with the current version, such as_ `/v0`.~

Implementation details for the **receiving** side ("holder"):

- ~`POST /holder/offers`~ `GET /openid4vci/offers` (_public_): this is the `credential_offer_endpoint` (as [defined here](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-13.html#section-4.1-1)), it accepts a `credential_offer` (by value) or a `credential_offer_uri` (by reference) which should be resolved. It should emit a `CredentialOfferReceived` event.
- `GET /v0/holder/offers`: returns all currently pending offers that show the actual credential to be received if accepted
- `POST /v0/holder/offers/{offer_id}/accept`: accepts the offer, which should trigger actually fetching the credential, `CredentialOfferAccepted` should be emitted
- `POST /v0/holder/offers/{offer_id}/reject`: rejects the offer, removes it from the list, (optional) inform the offering party about the rejection, event of type `CredentialOfferRejected` should be emitted

Implementation details for the **sending** side ("issuer"):

- ~`POST /offers/post`~ `POST /offers/send`: works the same way as the already existing `/offers` endpoint, but issues a `POST` request containing the offer to a target endpoint. Emits an `CredentialOfferSent` event. The required body should look as follows:

```diff
{
-  "openid-credential-offer": "openid-credential-offer://?credential_offer_uri=https%3A%2F%2Fserver%2Eexample%2Ec[...]"
+  "offer_id": "c77feaa4-eb7e-4ed4-bcbf-19358f2e98fb"
  "target_url": "https://example.org/holder/offers"
}
```

The `target_url` can be considered to be shared out-of-band between the two parties.

#### ~Authorization~

~Additionally, the `/holder/offers` endpoint should have some form of access control since it has to be publicly available to fulfill its purpose. Trust lists could be used to only allow verified parties to make offers. A more light-weight approach would be an "offer token" that the **holder** creates beforehand and shares it with the **issuer** out-of-band. The **issuer** has to include the token when it makes the offer. Every token can only be used _once_. This prevents spam and reduces the attack surface. An implementation could look as follows:~

~Implementation details for the **receiving** side ("holder"):~

- ~`POST /holder/offers/new`: creates a new "empty" offer. Should return a generated secure random token string. Only the fields `id` and `token` should be populated. `OfferTokenCreated` should be emitted.~
- ~update `POST /holder/offers` (_public_): should now accept an `application/json` payload that contains: `openid-credential-offer` and `token`. Provided `token` needs to be checked against all current `token`s (created throught the `/new` endpoint). If there's a match, add the `openid-credential-offer` to the offer, else return `401 Unauthorized`. An event `OfferTokenRedeemed` should be emitted. The `token` must now be treated as "burned" and removed from any offer instance (possibly set to `None`). Further calls, even with the same data, should also return a `401 Unauthorized`.~

### Motivation

Allow UniCore to receive credentials for its own identity and fulfill its "holder" role.

### Resources

n/a

### LINKED ISSUE #111: Authorized credential offers
### Description

In #109, credential offers between two UniCore instances is introduced. It requires a publicly available endpoint that can be called without authorization to receive credential offers. This is not necessarily insecure since the offer still needs to be manually confirmed to receive the credentials, but it allows spamming and unwanted offers.

In the follow, an _optional_ authorization extension is proposed:

- The `/holder/offers` endpoint should have some form of access control since it has to be publicly available to fulfill its purpose. Trust lists could be used to only allow verified parties to make offers. A more light-weight approach would be a "one-time offer token" that the **holder** creates beforehand and shares it with the **issuer** out-of-band. The **issuer** has to include the token when it makes the offer. Every token can only be used _once_. This prevents spam and reduces the attack surface. An implementation could look as follows:

Implementation details for the **receiving** side ("holder"):

- `POST /holder/offers/new`: creates a new "empty" offer. Should return a generated secure random token string. Only the fields `id` and `token` should be populated. `OfferTokenCreated` should be emitted.
- update `POST /holder/offers` (_public_): should now accept an `application/json` payload that contains: `openid-credential-offer` and `token`. Provided `token` needs to be checked against all current `token`s (created throught the `/new` endpoint). If there's a match, add the `openid-credential-offer` to the offer, else return `401 Unauthorized`. An event `OfferTokenRedeemed` should be emitted. The `token` must now be treated as "burned" and removed from any offer instance (possibly set to `None`). Further calls, even with the same data, should also return a `401 Unauthorized`.

### Motivation

Enhance credential distribution between UniCore instances

### Resources

_No response_

### To-do List

n/a

nanderstabel:
# Description of change
This PR adds Holder functionality to UniCore. 

* UniCore is now able to both send Credential Offers to a `target_url` as well as well as listen for incoming Credential Offers in order to (after confirmation) Accept or Reject the offer.
* Refactoring of `agent_api_rest`
* Abstracted Custom Query code into `agent_shared`
* Refactored test_utils across sub modules for test re-usability utilizing `rstest`

fixes https://github.com/impierce/ssi-agent/issues/111

Tested through new and existing (adjusted) unit tests as well as updated Postman Collection.

How to test this flow manually using the Postman Collection:
1. (Issuer): Hit the Issuance/w3c_vc_credentials request
2. (Issuer): Hit the Issuance/offers request
3. (Issuer): Hit the Issuance/offers_send request (this will send the Credential Offer to UniCore's new Holder module)
4. (Holder): Hit the Holder/offers request
    - (Holder): Hit the Holder/offers_accept request:
      - (Holder): Hit the Holder/credentials request to see the received Credential
    - (Holder): Or hit the Holder/offers_reject request:


agent_issuance/Cargo.toml
```diff
@@ -30,16 +30,25 @@ tracing.workspace = true
...
 tracing-test.workspace = true
 async-std = { version = "1.5", features = ["attributes", "tokio1"] }
-rstest.workspace = true
 
 [features]
-test_utils = []
+test_utils = [
+    "dep:lazy_static",
+    "dep:rstest",
+    "dep:once_cell"
```
daniel-mader: Should be sorted alphabetically

agent_secret_manager/src/service.rs
```diff
@@ -0,0 +1,20 @@
+use std::sync::Arc;
+
+/// Conventience trait for Services like `IssuanceServices`, `HolderServices`, and `VerifierServices`.
```
daniel-mader: Typo: `Conventience` => `Convenience`

agent_shared/src/config.rs
```diff
@@ -195,6 +226,12 @@ pub enum SupportedDidMethod {
     IotaRms,
 }
 
+impl From<SupportedDidMethod> for SubjectSyntaxType {
+    fn from(val: SupportedDidMethod) -> Self {
+        SubjectSyntaxType::try_from(val.to_string().as_str()).expect("convertion into `SubjectSyntaxType` failed")
```
daniel-mader: Typo: `convertion` => `conversion`

agent_api_rest/src/holder/holder/offers/accept.rs
```diff
@@ -0,0 +1,73 @@
...
+use agent_shared::handlers::{command_handler, query_handler};
+use axum::{
+    extract::{Path, State},
+    response::{IntoResponse, Response},
+};
+use hyper::StatusCode;
+
+#[axum_macros::debug_handler]
+pub(crate) async fn accept(State(state): State<HolderState>, Path(offer_id): Path<String>) -> Response {
+    // TODO: General note that also applies to other endpoints. Currently we are using Application Layer logic in the
```
daniel-mader: I had a hard time understanding this comment. I have suggestions improving the readability:
- `applies to other endpoints. Currently` => `applies to other endpoints: currently`
- `Furthermore, the to be implemented Application Layer` => `Furthermore, the Application Layer (not implemented yet)`

agent_api_rest/src/holder/holder/offers/reject.rs
```diff
@@ -0,0 +1,24 @@
...
+
+#[axum_macros::debug_handler]
+pub(crate) async fn reject(State(state): State<HolderState>, Path(offer_id): Path<String>) -> Response {
+    let command = OfferCommand::RejectCredentialOffer {
+        offer_id: offer_id.clone(),
+    };
+
+    // Remove the Credential Offer from the state.
+    if command_handler(&offer_id, &state.command.offer, command).await.is_err() {
+        // TODO: add better Error responses. This needs to be done properly in all endpoints once
```
daniel-mader, line 17: Maybe this issue can be used as a reference here: https://github.com/impierce/ssi-agent/issues/40

agent_api_rest/src/holder/mod.rs
```diff
@@ -0,0 +1,27 @@
+// TODO: further refactor the API's folder structure to reflect the API's routes.
```
daniel-mader, line 1: Could there be conflicts or inconsistencies if we follow that strategy (reflecting API routes with folder structure)? It's not uncommon, but does it also apply to this project?

agent_holder/src/offer/aggregate.rs
```diff
@@ -0,0 +1,485 @@
...
+                        .collect();
+
+                Ok(vec![CredentialOfferReceived {
+                    offer_id,
+                    credential_offer: Box::new(credential_offer),
+                    credential_configurations,
+                }])
+            }
+            AcceptCredentialOffer { offer_id } => {
+                // TODO: should we 'do nothing' or log a `warn!` message instead of returning an error?
```
daniel-mader: If this cannot happen from a business logic perspective, it's a "logical error" and should be logged and treated as any other error. Can we recover from it? Delete the offer? Auto-reject it?

daniel-mader (commented):
Looks like a well-functioning first iteration of the "holder role" 👍 I could not fully assess all technical refactorings, paired with new features - but I like the comments and tests to assert functionality and prevent bugs.

agent_api_rest/src/holder/mod.rs
nanderstabel: We can come up with our own logic and stay consistent to that. I think a nice structure would be to keep all the routes split between domain (issuance, holder etc) and another split between the UniCore API and oid4vc/well-known endpoints. 

Also, how do we structure functions for different methods (GET, POST) to the same endpoints? Do we put them in different files, or do we just name them differently, e.g: `post_credentials`, `get_credentials`.

If there is a conventional way of structuring this I think that can increase consistency. Not sure how there would be conflicts if we do it like this?

agent_holder/src/offer/aggregate.rs
daniel-mader: We agreed to process the offer again, even if it is not in the "pending" state. A user can accept an offer multiple times. A warning is still always logged. If there is no actual business case for accepting an offer multiple times, this part of the code can throw an error instead and abort the issuance.

agent_api_rest/src/holder/mod.rs
nanderstabel: I opened the following issue for this: https://github.com/impierce/ssi-agent/issues/120

daniel-mader (approved):
Awesome! All three major roles of SSI are now implemented in UniCore! 🥳

PR #135, opened 2024-10-09
feat: return collections as arrays, generic primary identifiers

### LINKED ISSUE #124: Change all `All<aggregate-type>View` formats from `HashMap` to `Vec`
### Description

For example this struct:
```rust

#[derive(Debug, Default, Serialize, Deserialize, Clone)]
pub struct AllReceivedOffersView {
    #[serde(flatten)]
    pub offers: HashMap<String, ReceivedOfferView>,
}

impl View<Offer> for AllReceivedOffersView {
    fn update(&mut self, event: &EventEnvelope<Offer>) {
        self.offers
            // Get the entry for the aggregate_id
            .entry(event.aggregate_id.clone())
            // or insert a new one if it doesn't exist
            .or_default()
            // update the view with the event
            .update(event);
    }
}
```

Needs to be re-written to something like (pseudo-code):
```rust

#[derive(Debug, Default, Serialize, Deserialize, Clone)]
pub struct AllReceivedOffersView {
    pub offers: Vec<ReceivedOfferView>,
}

impl View<Offer> for AllReceivedOffersView {
    fn update(&mut self, event: &EventEnvelope<Offer>) {
        self.offers.push(event)
    }
}
```

To find all structs that need to be adjusted search for this code snippet (there should be 4):
```rust
            // Get the entry for the aggregate_id
            .entry(event.aggregate_id.clone())
            // or insert a new one if it doesn't exist
            .or_default()
            // update the view with the event
            .update(event);
```

Important is that for each Vector item type, there needs to be a `..._id` field present. For example in the example above, the ID field needs to be added to `ReceivedOfferView`. In turn, this means that there needs to be an ID field added to `Offer` as well.

The end goal is that for all endpoints (there should be 4) that return All Credentials/Offers, the response type must change from an object `{}` to a list `[]` where all items in the list contain an ID field.

### Motivation

Better API Response types.

### Resources

n/a

### To-do List

- [ ] ..

nanderstabel:
# Description of change
As title. 

Also, changed some variable names to be internally consistent (`offer_id` --> `received_offer_id`)

## BREAKING CHANGE
BREAKING CHANGE: return collection as an array of objects instead of a map, rename primary identifier

The response types for the following endpoints have been changed from a map of objects to an array of objects:

- `/v0/holder/credentials`
- `/v0/holder/offers`
- `/v0/holder/presentations`
- `/v0/connections`
- `/v0/services`
- `/v0/credentials`
- `/v0/offers`

### Before
```
{
  {
    "credential_id": "bf766ebe-0a52-4531-ac24-88af1b854b55"
    ...
  }
}
```

### After
```
[
  {
    "id": "bf766ebe-0a52-4531-ac24-88af1b854b55"
    ...
  }
]
```

### Migration guide
- Parse responses as lists instead of objects
- Use the generic `id` property instead of having to search the response for the identifier

- #124
- #148 

Updated Postman Collection


agent_identity/src/service/aggregate.rs
```diff
@@ -32,7 +32,7 @@ pub enum ServiceResource {
 
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct Service {
-    pub id: String,
+    pub service_id: String,
```
daniel-mader, line 36: I think this change is not required as this carries enough context already:
```
pub struct Service {
    pub id: String
    ...
}
```

agent_issuance/src/credential/command.rs
```diff
@@ -7,13 +7,16 @@ use super::entity::Data;
 #[serde(untagged)]
 pub enum CredentialCommand {
     CreateUnsignedCredential {
+        credential_id: String,
```
daniel-mader, line 10: Is the caller now _required_ to pass in an identifier for the credential?

agent_identity/src/service/aggregate.rs
daniel-mader: If you want to keep it explicit and consistent throughout the entire code, I would be okay with the change.

agent_issuance/src/credential/command.rs
nanderstabel: Yes when invoking this Command you will need to add this identifier explicitly since there is no other way to add an aggregates ID (not that I am aware of at least).

agent_identity/src/service/aggregate.rs
nanderstabel: Changed all aggregate ID fields to `id` :ok_hand:

agent_issuance/src/credential/command.rs
daniel-mader: What about generating it inside the command handler?

nanderstabel: This is not possible with the `cqrs` framework we're using because the underlying `CqrsFramework::execute` method requires the aggregate ID to already exist before executing the Command: https://github.com/serverlesstechnology/cqrs/blob/main/src/cqrs.rs#L124-L131

nanderstabel: The underlying `CqrsFramework::execute` method requires the aggregate ID to be there before the actual Command is executed, so the aggregate ID needs to be generated 'outside' of the command handler. See: https://github.com/serverlesstechnology/cqrs/blob/main/src/cqrs.rs#L124-L131

daniel-mader: Ah, my main concern was that the caller has to generate and pass in the ID themselves, but I just found that it's still generated inside the application (`agent_api_rest`). 👍

daniel-mader (approved):

PR #158, opened 2024-12-16
feat: add `/version` and `/info` metadata endpoints

daniel-mader:
# Description of change

Two new endpoints are introduced:
- `/version`: returns the version (taken from the git tag) and the git commit hash (short)
- `/info`: returns the name of the application ("UniCore"), all values from `/version`, as well as the container uptime and the timestamp when the container was built

The capabilities are described in [docs/deployment/monitoring.md](https://github.com/impierce/ssi-agent/blob/feat/version/docs/deployment/monitoring.md)

This Docker image was built containing the changes: `impiercetechnologies/ssi-agent:sha-42d092c`

n/a

n/a


.github/workflows/build-push-docker.yaml
```diff
@@ -55,18 +55,22 @@ jobs:
           platforms: ${{ matrix.platform }}
           labels: ${{ steps.meta.outputs.labels }}
           outputs: type=image,name=${{ env.REGISTRY_IMAGE }},push-by-digest=true,name-canonical=true,push=true
+          build-args: |
+            APP_VERSION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.version'] }}
```
nanderstabel, line 65: So for the image that you created from this branch the value for `APP_VERSION` is `feat-version` so it seems like it was derived from the branch name. Does 'it' automatically understand that the version should be `1.0.0-beta.6` once a new UniCore version is released?

agent_application/build.rs
```diff
@@ -0,0 +1,42 @@
...
+        writer,
+        "pub const DOCKER_BUILD_TIMESTAMP: Option<&str> = {:?};",
+        docker_build_timestamp
+    )?;
+    writeln!(writer, "#[allow(dead_code)]")?;
+    writeln!(
+        writer,
+        "pub const APP_RELEASE_CHANNEL: Option<&str> = {:?};",
+        app_release_channel
+    )?;
```
nanderstabel: You could consider using literal raw string notation to make this (slightly) more readable like this:
```rust
    writeln!(
        writer,
        r#"// THIS FILE IS GENERATED BY A BUILD SCRIPT
pub const APP_VERSION: Option<&str> = {app_version:?};
pub const GIT_COMMIT_HASH: Option<&str> = {git_commit_hash:?};
#[allow(dead_code)]
pub const DOCKER_BUILD_TIMESTAMP: Option<&str> = {docker_build_timestamp:?};
#[allow(dead_code)]
pub const APP_RELEASE_CHANNEL: Option<&str> = {app_release_channel:?};
"#
    )?;
```

agent_application/src/metadata/info.rs
```diff
@@ -0,0 +1,50 @@
...
+        git_commit_hash: GIT_COMMIT_HASH
+            .filter(|s| !s.trim().is_empty())
+            .map(|s| s.to_string().chars().take(7).collect()),
+        release_channel: APP_RELEASE_CHANNEL
+            .filter(|s| !s.trim().is_empty())
+            .map(|s| s.to_string()),
+        docker_build_timestamp: DOCKER_BUILD_TIMESTAMP
+            .filter(|s| !s.trim().is_empty())
+            .map(|s| s.to_string()),
+        uptime: uptime_human_readable,
```
nanderstabel: Why do these values need to trimmed?

agent_application/src/metadata/info.rs
```diff
@@ -0,0 +1,50 @@
...
+    version: Option<String>,
+    /// The git commit hash from which the application was built.
+    git_commit_hash: Option<String>,
+    /// The release channel of the application. Possible values are: `stable`, `next`, `beta`, `canary`.
+    release_channel: Option<String>, // TODO: mapping ok? stable/latest (main), next, beta, canary (alpha)
+    /// The timestamp when the Docker image was built.
+    docker_build_timestamp: Option<String>,
+    /// The current uptime of the application in a human-friendly format.
+    uptime: String,
+}
```
nanderstabel: Nit:

You could choose to reuse your `Version` struct here so that potential future changes will be automatically be reflected both for the `/version` endpoint as well as here in the `/info` endpoint (if that's desirable):
```rust
#[skip_serializing_none]
#[derive(Serialize)]
pub struct Info {
    #[serde(flatten)]
    version: super::version::Version,
    /// The release channel of the application. Possible values are: `stable`, `next`, `beta`, `canary`.
    release_channel: Option<String>, // TODO: mapping ok? stable/latest (main), next, beta, canary (alpha)
    /// The timestamp when the Docker image was built.
    docker_build_timestamp: Option<String>,
    /// The current uptime of the application in a human-friendly format.
    uptime: String,
}
```

agent_application/src/metadata/version.rs
```diff
@@ -0,0 +1,23 @@
...
+    version: Option<String>,
+    /// The git commit hash from which the application was built.
+    git_commit_hash: Option<String>,
+}
+
+/// Returns the `version` and the `git_commit_hash` of the application.
+pub async fn version() -> Json<Version> {
+    let version = Version {
+        version: APP_VERSION.map(|s| s.to_string()),
+        git_commit_hash: GIT_COMMIT_HASH.map(|s| s.to_string().chars().take(7).collect()),
```
nanderstabel: Why does `git_commit_hash` in `fn info()` needs to be trimmed, but here it doesn't?

docs/deployment/monitoring.md
```diff
@@ -0,0 +1,27 @@
...
+
+- `/healthz`: Liveness probe
+<!-- - `/readyz`: Readiness probe -->
+
+## Metadata
+
+The following endpoints provide metadata about the service itself and allow for more detailed deployment monitoring:
+
+- `/version`: Returns the version and the git commit hash
+- `/info`: Returns the version, the git commit hash, a Docker build timestamp and the container uptime
```
nanderstabel, line 17: I'm just wondering, so these endpoints seem to be semi-conventional, but there doesn't seem to be any specification that specifies how they should be implemented (afaik?). So did you come up yourself with this format:
```json
{
    "version": "feat-version",
    "git_commit_hash": "18f26e3",
    "docker_build_timestamp": "2025-03-14T12:46:57.821Z",
    "uptime": "0 days, 00:29:38"
}
```
Should/could we follow any convention here? (I don't know I'm just interested :))

nanderstabel (commented):
Looks good! I just left some minor comments/questions :+1:

.github/workflows/build-push-docker.yaml
daniel-mader: Yes, I inspected the meta information during "dirty" builds and proper release builds. If a git tag is present, it uses that one. I can also adjust the metadata generation and generate a `sha-<commit>` tag as well, if you prefer.

agent_application/src/metadata/info.rs
daniel-mader: Because
```
"".is_empty();          <== true
"   ".is_empty();       <== false
```

I am aware that those are not externally-provided or user-defined values (they come from the CI), so the changes are highly unliked that this extra check is ever needed, but it's just generally a good idea to `trim()` before checking `is_empty()` if you're interested in _truly blank_ values.

agent_application/src/metadata/version.rs
daniel-mader: You're right, it should be trimmed in both cases.

docs/deployment/monitoring.md
daniel-mader: There is no _actual_ convention for it, as far as I know. Spring Boot (a widely-used Java framework) has a [similar feature](https://www.baeldung.com/spring-boot-info-actuator-custom), but the keys are non-conventional. You can basically put anything there that you think could be interesting for a developer/operator.

daniel-mader: We could also consider `revision` instead of `git_commit_hash`.

agent_application/src/metadata/info.rs
nanderstabel: Ok and is this (although very unlikely) something you only see that could happen for values that are supposed to be blank?

Could actual values also contain whitespaces, for example `APP_VERSION = Some("version\n")`?

If so then maybe it is better to first `map` and then `filter` e.g.:

```rust
#[test]
fn test() {
    let git_commit_hash = Some("\t\t58dw34n89572t3n".to_string());

    let current = git_commit_hash
        .clone()
        .filter(|s| !s.trim().is_empty())
        .map(|s| s.to_string().chars().take(7).collect::<String>());

    // Still contains whitespace
    assert_eq!(current, Some("\t\t58dw3".to_string()));

    let suggested = git_commit_hash
        .map(|s| s.trim().chars().take(7).collect::<String>())
        .filter(|s| s.len() == 7);

    // Whitespace removed
    assert_eq!(suggested, Some("58dw34n".to_string()));
}
```

docs/deployment/monitoring.md
nanderstabel: Yeah `revision` is nice, but `git_commit_hash` works as well imo :+1:

.github/workflows/build-push-docker.yaml
nanderstabel: Yeah actually I think it would be nice to have both!

docs/deployment/monitoring.md
daniel-mader: The whole point of this endpoint is to give an operator insight into what version of the application is running on that URL. So it should be human-readable and not too technical and/or nested. While `rev` or `revision` would be more consistent with the style cargo dependencies are defined, it's less apparent to an operator (that probably doesn't even know Rust). Since we're in the lead and there's no actual convention for this endpoint, I'm fine with both `revision` (more technical) or `git_commit_hash` (more human-readable).

.github/workflows/build-push-docker.yaml
daniel-mader: The tags have been updated and builds from untagged commits now look like so:

```
GET /version

{
  "version": "sha-9c10f4d"
}
```

nanderstabel: Great, nice!

nanderstabel (approved):

PR #166, opened 2025-02-24
feat: use `credential_offer_uri` by default during credential issuance

### LINKED ISSUE #165: Add the `credential_offer_uri` to `agent_issuance`
### Description

According to the OpenID4VCI specification, issuers can send their [Credential Offer by Value](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-13.html#name-sending-credential-offer-by) or [by Reference](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-13.html#name-sending-credential-offer-by-).

The `CreateCredentialOffer` handler in the `agent_issuance` module (`agent_issuance/src/offer/aggregate.rs`) should be updated so that, by default, the offer is created as `CredentialOffer::CredentialOfferUri` rather than `CredentialOffer::CredentialOffer`. Only when a new configuration variable (to be implemented) is enabled should the latter be used.

### Motivation

If a Credential Offer is very large, sending it by Value can cause issues when rendering it as a QR code. Therefore, it is preferable to send the Credential Offer by Reference to keep the QR code small.

Currently, UniCore does not support Credential Offers by Reference, so this functionality needs to be implemented.

### Resources

https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-13.html#name-sending-credential-offer-by-

### To-do List

- [ ] Add a new `credential_offer_uri` endpoint to `agent_api_rest`
- [ ] Add a new configuration variable (see `config.rs`)
- [ ] Update the handler for `CreateCredentialOffer`

coplat:
# Description of change
Updates `CreateCredentialOffer` handler to default offer creations to by reference as opposed to by value. 

Fixes issue #165

cargo test, some tests still failing - being worked on. 

- [x ] I have followed the contribution guidelines for this project


agent_api_rest/src/issuance/credential_issuer/credential.rs
```diff
@@ -266,10 +266,10 @@ mod tests {
...
     #[rstest]
     #[case::without_external_server(false, false, 0)]
-    #[case::with_external_server(true, false, 0)]
-    #[case::with_external_server_and_self_signed_credential(true, true, 0)]
-    #[should_panic(expected = "assertion `left == right` failed\n  left: 500\n right: 200")]
-    #[case::should_panic_due_to_timout(true, false, DEFAULT_EXTERNAL_SERVER_RESPONSE_TIMEOUT_MS + 100)]
+    // #[case::with_external_server(true, false, 0)]
+    // #[case::with_external_server_and_self_signed_credential(true, true, 0)]
+    // #[should_panic(expected = "assertion `left == right` failed\n  left: 500\n right: 200")]
+    // #[case::should_panic_due_to_timout(true, false, DEFAULT_EXTERNAL_SERVER_RESPONSE_TIMEOUT_MS + 100)]
```
nanderstabel: All these test as well as any other failing test in `agent_api_rest` can be fixed simply by setting the new config variable to true: `credential_offer_by_value_enabled: true` in the `test-config.yaml` file. 

So no other alterations to any of the tests in `agent_api_rest` are needed since setting the config variable to `true` is basically already the status quo when it comes to these tests :)

agent_issuance/src/offer/aggregate.rs
nanderstabel, line 1: The `test_create_offer` test in this file is failing because it uses `Default::default()` as a value for `offer_id`.

The cleanest way to solve this is to include the `offer_id` into the test and use it as the value for every `offer_id` variable, e.g.:
```rust
    #[rstest]
    #[serial_test::serial]
    async fn test_create_offer(
        offer_id: String,
        #[future(awt)] pre_authorized_code: String,
        #[future(awt)] access_token: String,
        credential_issuer_metadata: Box<CredentialIssuerMetadata>,
        #[future(awt)] credential_offer: CredentialOffer,
        #[future(awt)] credential_offer_uri: CredentialOffer,
    ) {
        OfferTestFramework::with(Service::default())
            .given_no_previous_events()
            .when(OfferCommand::CreateCredentialOffer {
                offer_id: offer_id.clone(),
                credential_issuer_metadata,
            })
            .then_expect_events(vec![OfferEvent::CredentialOfferCreated {
                offer_id,
                credential_offer,
                credential_offer_uri,
                pre_authorized_code,
                access_token,
                status: Status::Created,
            }]);
    }
```

Even though the other tests in this file are not failing, it is probably still best to do the same for all the other tests that have `offer_id: Default::default()` as well. :+1:

agent_api_rest/src/issuance/credential_issuer/credential.rs
```diff
@@ -283,6 +283,7 @@ mod tests {
 
             let target_url = format!("{}/ssi-events-subscriber", &external_server.uri());
 
+            set_config().credential_offer_by_value_enabled = Some(true);
```
nanderstabel: Setting the variable like this won't be necessary after you set it directly in the `test-config.yaml` file (see comment above).

agent_issuance/src/offer/aggregate.rs
```diff
@@ -88,8 +91,17 @@ impl Aggregate for Offer {
                     }),
                 }));
 
+                let credential_offer_uri = CredentialOffer::CredentialOfferUri(
+                    Url::parse(&format!(
+                        "{}/credential-offer/{}",
+                        credential_issuer_metadata.credential_issuer, offer_id
+                    ))
```
nanderstabel: This would result in a URL like this:
`https://example.com//credential-offer/offer_id` (has double `/` in front of the `credential-offer` path segment)
instead of what you actually want:
`https://example.com/credential-offer/offer_id`

I've ran into similar issues myself when parsing formatted strings and now I believe it is a bit safer to use the `Url::join` method to add path segments to `Url`s :) : 
```rust
                let credential_offer_uri = CredentialOffer::CredentialOfferUri(
                    credential_issuer
                        .join("credential-offer/")
                        .and_then(|url| url.join("offer_id"))
                        .map_err(|e| InvalidUrlError(e.to_string()))?,
                );
```

agent_issuance/src/offer/aggregate.rs
```diff
@@ -606,11 +651,30 @@ pub mod test_utils {
...
+    #[fixture]
+    pub async fn credential_offer_uri(
+        credential_issuer_metadata: Box<CredentialIssuerMetadata>,
+        offer_id: String,
+    ) -> CredentialOffer {
+        CredentialOffer::CredentialOfferUri(
+            Url::parse(&format!(
+                "{}/credential-offer/{}",
+                credential_issuer_metadata.credential_issuer, offer_id
+            ))
```
nanderstabel: See comment about `Url::join` above

agent_issuance/src/offer/aggregate.rs
```diff
@@ -103,24 +115,42 @@ impl Aggregate for Offer {
...
+                        .as_ref()
+                        .ok_or(MissingCredentialOfferError)?
+                        .to_string()
+                } else {
+                    // default: use credential_offer_uri
+                    self.credential_offer_uri
+                        .as_ref()
+                        .ok_or(MissingCredentialOfferError)?
+                        .to_string()
+                };
```
nanderstabel: Makes total sense that you the same method to construct the `form_url_encoded_credential_offer` here as well! However, seeing this makes me think that 'recreating' the `form_url_encoded_credential_offer` here was not really a good idea in the first place. 

We should probably discuss this further in a call, but my idea improve this is as follows:
1. We get rid of the `CreateFormUrlEncodedCredentialOffer` command
2. We keep the `FormUrlEncodedCredentialOfferCreated` event the way you updated it, but we add it to the `CreateCredentialOffer` command. So `CreateCredentialOffer` will result in two events: `CredentialOfferCreated` and `FormUrlEncodedCredentialOfferCreated`.
3. In `SendCredentialOffer` we can simply assume that `self.form_url_encoded_credential_offer` contains the `form_url_encoded_credential_offer` string.

nanderstabel (changes requested):
A bunch of tests related to `IotaRms` are currently failing because of the discontinuation of the `did:iota:rms` DID Method (see https://github.com/impierce/ssi-agent/issues/167) so this is unrelated to your PR. I will fix that problem in a separate PR so this problem will resolve itself eventually :+1:

agent_issuance/src/offer/aggregate.rs
coplat: In this case, could we also simplify the `SendCredentialOffer` command to simply use the url encoded offer? As opposed to needing to check config again.

nanderstabel:
More details for updating the Postman Collection:

Current test script flow in the Postman Collection (Credential Offer by Value)
1. `POST Issuance/w3c_vc_credentials`
2. `POST Issuance/offers`
  * Its Tests script extracts the `pre-authorized_code` from the Response and stores it in the `PRE_AUTHORIZED_CODE` environment variable.
3. `GET oid4vci/oauth-authorization-server`, its Tests script:
  * Stores the `issuer` in the `ISSUER` environment variable.
  * Stores the `token_endpoint` in the `TOKEN_ENDPOINT` environment variable.
4. `GET oid4vci/openid-credential-issuer`
5. `POST oid4vci/token` (uses the `PRE_AUTHORIZED_CODE` environment variable)
    * Its Tests script extracts the `access_token` from the Response and stores it in the `ACCESS_TOKEN` environment variable.
6. `POST oid4vci/credential` (uses the `ACCESS_TOKEN` environment variable)

**Desired test script flow**:
1. Same as above
2. The test script should extract (and decode) the value for `credential_offer_uri` and strore it in a new environment variable called `CREDENTIAL_OFFER_URI`.
3. A new request should be added: `GET oid4vci/Get Credential Offer by Value`
  * Its URL should be `{{CREDENTIAL_OFFER_URI}}`
  * Its Tests script should extract the `pre-authorized_code` from the Response and store it in the `PRE_AUTHORIZED_CODE` environment variable (note that the response is not a URL encoded string now, but alrady a simple JSON object)
4. Same as step 3 above
5. Same as step 4 above
6. Same as step 5 above
7. Same as step 6 above

---

* The new `CREDENTIAL_OFFER_URI` environment variable should be added first to the `ssi-agent` Collection (in the 'Variables' tab)
* So that means that only step 2 needs to be updated, and a new request with Test script should be added.

coplat:
Hello

Since defaulting to offers by reference, we recently had to add an additional offers handler function (`fn offers_by_reference`) to manage these. I've just added a commit to merge these two functions into one main `offers` function. Let me know how it looks!

agent_secret_manager/tests/res/temp.stronghold
nanderstabel, line 1: It seems that this has been accidentally added :) (should be removed)

agent_issuance/src/offer/aggregate.rs
```diff
@@ -88,13 +90,38 @@ impl Aggregate for Offer {
...
+                        .and_then(|url| url.join(&offer_id))
+                        .map_err(|e| InvalidUrlError(e.to_string()))?,
+                );
+
+                let credential_offer_by_value_enabled = config().credential_offer_by_value_enabled.unwrap_or_default();
+
+                let form_url_encoded_credential_offer = if credential_offer_by_value_enabled {
+                    credential_offer.to_string()
+                } else {
+                    //this makes that by default, credenital_offer_uri is used.
```
nanderstabel: I think comment can be removed. This code block is clear enough by itself and this command handler should not really be concerned with whether `true` or `false` is the default value for `form_url_encoded_credential_offer`.

agent_issuance/src/offer/aggregate.rs
```diff
@@ -103,27 +130,12 @@ impl Aggregate for Offer {
...
-            CreateFormUrlEncodedCredentialOffer { offer_id } => Ok(vec![FormUrlEncodedCredentialOfferCreated {
-                offer_id,
-                form_url_encoded_credential_offer: self
-                    .credential_offer
-                    .as_ref()
-                    .ok_or(MissingCredentialOfferError)?
-                    .to_string(),
-                status: Status::Pending,
-            }]),
+
```
nanderstabel: A bit nit-picky :p but let's remove this empty line to stay consistent throughout the code :+1:

nanderstabel (changes requested):

agent_issuance/src/offer/aggregate.rs
coplat: removed :)

agent_issuance/src/offer/aggregate.rs
coplat: 👍

agent_api_rest/src/issuance/credential_issuer/token.rs
```diff
@@ -73,7 +73,11 @@ pub mod tests {
     use oid4vci::token_response::TokenResponse;
     use tower::Service as _;
 
-    pub async fn token(app: &mut Router, pre_authorized_code: String) -> String {
+    pub async fn token(app: &mut Router, pre_authorized_code: Option<String>) -> Option<String> {
+        let pre_authorized_code = match pre_authorized_code {
+            Some(code) => code,
+            None => return None,
+        };
```
nanderstabel: This `token` function should stay the way it was since it should always expect an `pre_authorized_code` and it should always return the `token_response.access_token` (unless it panicked before it could return anything).

agent_api_rest/src/issuance/credential_issuer/token.rs
```diff
@@ -100,19 +104,20 @@ pub mod tests {
...
-
         let mut app = router(issuance_state);
 
         credentials(&mut app).await;
-        let pre_authorized_code = offers(&mut app).await;
+        let pre_authorized_code: String = offers(&mut app).await.unwrap();
 
-        let _access_token = token(&mut app, pre_authorized_code).await;
+        let _access_token = token(&mut app, Some(pre_authorized_code)).await;
+        assert!(_access_token.is_some());
```
nanderstabel: This won't be necessary since the `token` function should always return a `String` (instead of `Option<String>`)

agent_api_rest/src/issuance/credential_issuer/token.rs
```diff
@@ -100,19 +104,20 @@ pub mod tests {
         assert_eq!(token_response.token_type, "bearer");
         assert!(token_response.c_nonce.is_some());
 
-        token_response.access_token
+        Some(token_response.access_token)
```
nanderstabel: This should return `token_response.access_token`

agent_api_rest/src/issuance/credential_issuer/credential.rs
```diff
@@ -319,14 +319,15 @@ mod tests {
         let pre_authorized_code = offers(&mut app).await;
```
nanderstabel: This should be changed to `let pre_authorized_code = offers(&mut app).await;`

agent_api_rest/src/issuance/credential_issuer/credential.rs
```diff
@@ -319,14 +319,15 @@ mod tests {
...
+        let access_token_value = access_token.unwrap_or_else(|| panic!("No access token available"));
 
         let response = app
             .oneshot(
                 Request::builder()
                     .method(http::Method::POST)
                     .uri("/openid4vci/credential")
                     .header(http::header::CONTENT_TYPE, mime::APPLICATION_JSON.as_ref())
-                    .header(http::header::AUTHORIZATION, format!("Bearer {}", access_token))
+                    .header(http::header::AUTHORIZATION, format!("Bearer {}", access_token_value))
```
nanderstabel: This should be changed back to the way it was since the `token` function should always return `String` (instead of `Option<String>`)

nanderstabel (changes requested):

agent_api_rest/src/issuance/credential_issuer/credential.rs
coplat: hm it already is as such

nanderstabel: Oops sorry I meant it should be changed to `let pre_authorized_code = offers(&mut app).await.unwrap();` :sweat_smile:

agent_api_rest/src/issuance/credential_issuer/token.rs
nanderstabel: The original function was `pub async fn token(app: &mut Router, pre_authorized_code: String) -> String`, so:
```rust
pre_authorized_code: Option<String>
```
should be changed back to:
```rust
pre_authorized_code: String
```

Which also means that this code won't be necessary anymore:
```rust
        let pre_authorized_code = match pre_authorized_code {
            Some(code) => code,
            None => return None,
        };
```

agent_api_rest/src/issuance/credential_issuer/token.rs
nanderstabel: So related to the comment above, this line:
```rust
let _access_token = token(&mut app, Some(pre_authorized_code)).await;
```

Should be changed to:
```rust
let _access_token = token(&mut app, pre_authorized_code).await;
```

because the `token` function should expect `pre_authorized_code: String` instead of `pre_authorized_code: Option<String>` :)

agent_api_rest/src/issuance/credential_issuer/credential.rs
coplat: ok made the changes. all should work now :)

nanderstabel (approved):
All good, great work! :partying_face:

PR #182, opened 2025-03-26
feat: introduce application profiles with sensible default configuration

daniel-mader:
# Description of change

- UniCore can now run in two "profiles": `production` _(default)_, `development` (<== set via env `UNICORE__PROFILE`)
- All configuration options are now optional and fallback to sensible defaults in the `development` profile
- This makes UniCore's configuration mostly self-contained and reduces initial DX friction
- Some configuration values that can not have a sensible default (such as passwords) are _required_ in `production`
- The current configuration can be inspected at `GET /v0/configuration`
  - an optional query parameter `?provisioned=true` can be used to filter for all values that were _explicitly_ provided through a config file or set via an environment variable
- A `POST` to `/v0/configuration` is not part of this change, but allows future implementations of runtime config changes (NOTE: provisioned config values are excluded from runtime updates for consistency reasons)

n/a

n/a


nanderstabel:
I've looked into the failing tests and there are 2.5 causes:
1.
`test_qr_code_scanned_handle_siopv2_authorization_request` and `test_qr_code_scanned_handle_oid4vp_authorization_request` fail because during both tests, the Authorization Response is send to http://example.com, however this domain doesn't exist anymore and therefore the tests both time out. So these failing tests are not introduced in this PR. I've validated that this was the issue by replacing the domain to http://example.org (http://example.test also doesn't exist anymore), you can check out the changes in the `fix/example-org` branch. This would be a quick fix, but relying on external domain is not a good idea like we've discussed before. Perhaps it is better to change the test domains to https://www.impierce.com/ (because this domain will live forever :wink:) since the only requirement for the tests is that it can send an HTTP request (they don't care about the responses, only whether the domain exists). Of course ultimately the best solution would be to test against a test UniCore instance.

2.
`test_get_state_create_new` and `test_locale_stays_unchanged_on_profile_creation` are failing because their corresponding fixture JSON files do not contain the newly introduced `biometrics_enabled` field in the `CreateNew` Action. So you can either add those fields to the corresponding fixtures, or probably better: add `#[serde(default)]` so that the missing field will default to `false`.
```rust
pub struct CreateNew {
    pub name: String,
    pub picture: String,
    pub theme: AppTheme,
    pub password: String,
    #[serde(default)]
    pub biometrics_enabled: bool,
}
```

2.5
`verifier_successfully_verifies_eddsa_signed_data` is still flaky because `test_credential_is_removed_from_appstate_and_from_stronghold_and_image_is_deleted` is not serial. This can easily be fixed by adding the `serial` attribute:
```rust
    #[tokio::test]
    #[serial_test::serial]
    async fn test_credential_is_removed_from_appstate_and_from_stronghold_and_image_is_deleted()
```

nanderstabel (dismissed):

nanderstabel (approved):

PR #186, opened 2025-04-01
feat: add `public_url`, rename `url` to `application_url`

### LINKED ISSUE #163: Redefine exposed URLs
### Description

UniCore uses the `UNICORE__URL` config value as a "self-aware URL" which it will put into all sorts of configurations that are made available to the outside (such as `.well-known` endpoints, as well as in `did:web`, etc.).
In some deployment scenarios, a "base path" is required with will be used as a prefix for all endpoints.

In some scenarios, the "public-facing" URL should be a completely different one than the "internal" URL (including a base path).

The following changes are proposed:

* `UNICORE__APPLICATION_URL`: rename from `UNICORE__URL`
* `UNICORE__BASE_PATH` (optional): _unchanged_
* `UNICORE__PUBLIC_URL` (optional): overwrites the value from `APPLICATION_URL` (including the `BASE_PATH` if set) for all publicly exposed URLs as described above

### Motivation

Support more deployments setups

### Resources

_No response_

### To-do List

see **Description**

### LINKED ISSUE #164: [Bug] `BASE_PATH` does not support multiple levels
### Description

Setting `UNICORE__BASE_PATH=/foo/bar` will only reflect as `http://<HOST>/foo` instead of `http://<HOST>/foo/bar`.

### Hardware Specification

_No response_

### Steps to Reproduce the Bug

n/a

### Expected Behaviour

n/a

### Actual Behaviour

n/a

### Errors

n/a

nanderstabel:
# Description of change
This PR introduces several configuration improvements and breaking changes to environment variables for UniCore:

Main Updates
* Fixed issues with (missing) trailing slashes
* Renamed `UNICORE__URL` to `UNICORE__APPLICATION_URL` for clarity and consistency.
* Added a new `UNICORE__PUBLIC_URL` environment variable. This value is intended for communication with clients and identity wallets and should be publicly accessible.
* Introduced new environment variables for advanced endpoint customization:
  * `UNICORE__TOKEN_ENDPOINT`
  * `UNICORE__CREDENTIAL_ENDPOINT`
  * `UNICORE__CREDENTIAL_OFFER_URI`
  * `UNICORE__REQUEST_URI`
  * `UNICORE__REDIRECT_URI`
* Documentation updated to clarify default behaviors and recommend using defaults in most setups.

Breaking Changes
* Removed `UNICORE__PORT` and `UNICORE__BASE_PATH` environment variables. Their functionality is now covered by `UNICORE__APPLICATION_URL`, which can include port and base path components as needed for internal use.
**Note**: Please review your deployment and configuration files to ensure compatibility with these changes.

**Blocked by**
- https://github.com/impierce/ssi-agent/pull/182

closes #163 
closes #164

Multiple unit tests added


agent_identity/src/document/aggregate.rs
```diff
@@ -706,7 +705,7 @@ pub mod test_utils {
             .type_("LinkedDomains")
             .service_endpoint(
                 ServiceEndpoint::from_json_value(json!({
-                    "origins": [config().url],
+                    "origins": [config().public_url.clone()],
```
daniel-mader: This is just test code, but shouldn't this be equivalent to the production code:

```
config().public_url.origin().clone()
                    ^^^^^^^^
```

just to make sure?

agent_shared/src/config/mod.rs
```diff
@@ -1198,4 +1311,87 @@ mod tests {
...
         );
     }
+
+    #[test]
+    #[serial]
+    fn test_public_url_defaults_to_application_url() {
+        temp_env::with_vars(
+            [
+                ("UNICORE__EVENT_STORE__CONNECTION_STRING", Some("postgresql://:test:")),
+                ("UNICORE__SECRET_MANAGER__STRONGHOLD_PASSWORD", Some("unsafe-password")),
```
daniel-mader: Are those env variables required in the test, even if `ApplicationProfile::Development` is used?

docs/configuration/CONFIGURATION.md
```diff
@@ -45,36 +45,176 @@ This overview is still a work in progress. You can refer to the [example.config.
 
 :::
 
-### URL
+### Application URL
 
-UniCore's base URL. This value is communicated to clients and identity wallets.
+UniCore's application URL. This value represents the self-aware URL of the application. It is used for internal
+communication and should not be exposed to clients or identity wallets.
```
daniel-mader, line 51: I think we should be a bit more precise here. Why _should_ it not be exposed? For security reasons? We can for example say: **"This is the URL that you can use for routing behind a gateway or reverse proxy"**. I also think we should add a reference to the "Public URL" in this description already to avoid confusion: **"This might differ from from the URL that UniCore should use for external communication with wallets and other services"**.

docs/configuration/CONFIGURATION.md
```diff
@@ -45,36 +45,176 @@ This overview is still a work in progress. You can refer to the [example.config.
...
+:::
+
+| Environment variable       | `config.yaml`     |
+| -------------------------- | ----------------- |
+| `UNICORE__APPLICATION_URL` | `application_url` |
+
+#### Example
+
+```yaml
+url: http://localhost:3033/my/base/path
```
daniel-mader: Shouldn't this be `application_url`?

docs/configuration/CONFIGURATION.md
```diff
@@ -45,36 +45,176 @@ This overview is still a work in progress. You can refer to the [example.config.
...
+:::
+
+| Environment variable  | `config.yaml` |
+| --------------------- | ------------- |
+| `UNICORE__PUBLIC_URL` | `public_url`  |
 
 #### Example
 
 ```yaml
 url: https://my-domain.example.test
```
daniel-mader: Shouldn't this be changed to `public_url`?

docs/configuration/CONFIGURATION.md
```diff
@@ -45,36 +45,176 @@ This overview is still a work in progress. You can refer to the [example.config.
...
+### Public URL
+
+UniCore's public URL. This value is communicated to clients and identity wallets and should be publicly accessible. When
+not set, it defaults to the value of `UNICORE__APPLICATION_URL`.
+
+<!-- TODO: In production, should we require that `UNICORE__PUBLIC_URL` is always explicitly set, rather than defaulting to the value of `UNICORE__APPLICATION_URL`? This would help prevent accidental exposure of internal URLs to clients. -->
+
+:::note
+
+The `UNICORE__PUBLIC_URL` may also include a path segment, which will be treated as the base path for all public endpoints. For example, if you set `UNICORE__PUBLIC_URL` to `https://my-domain.example.test/my/base/path`, then all public API endpoints will be served under `/my/base/path`.
```
daniel-mader, line 78: * We should add the disclaimer here as well that Domain Linkage is blocked, if there is a base path set.
* We should fail config validation entirely when there is a base path and domain linkage enabled.
* We could also block base paths altogether from the `PUBLIC_URL` and suggest other custom routing to instances behind a gateway/reverse proxy, for example based on specific custom headers instead of base paths.

docs/configuration/CONFIGURATION.md
```diff
@@ -45,36 +45,176 @@ This overview is still a work in progress. You can refer to the [example.config.
...
-| -------------------- | ------------- |
-| `UNICORE__PORT`      | `port`        |
+The OAuth2/OpenID Connect token endpoint. This endpoint is used by clients to exchange authorization codes for access tokens.
 
-#### Values
+:::note
+
+This variable is **optional**. By default, the `UNICORE__PUBLIC_URL` is used as the base, and the `/auth/token` segment is appended to form the token endpoint URL. You can completely override this default by explicitly setting the `UNICORE__TOKEN_ENDPOINT` variable or the `token_endpoint` config value.
+
+In most setups, the default value is recommended and usually the best choice.
```
daniel-mader, line 102: I suggest a different wording here: **For most setups, the default value is recommended and should only be changed if you know what you're doing.** This makes it clearer that there's a chance of potential misconfiguration, if you're just "playing around" with the config.

daniel-mader (changes requested):
Looks good, just some minor comments on the docs 👍

docs/configuration/CONFIGURATION.md
nanderstabel: yes indeed, I overlooked this one!

docs/configuration/CONFIGURATION.md
nanderstabel: As well as this one :+1:

agent_identity/src/document/aggregate.rs
nanderstabel: Agreed ==> fixed :ok_hand:

agent_shared/src/config/mod.rs
nanderstabel: indeed ==> removed :wastebasket:

docs/configuration/CONFIGURATION.md
nanderstabel: Agree with your suggestions :+1:

docs/configuration/CONFIGURATION.md
nanderstabel: Agree :+1:

docs/configuration/CONFIGURATION.md
nanderstabel: 1. agree
2. agree, although technically it would be possible that the reverse proxy would route everything from the base path to the corresponding routes in the UniCore instance, but add an exception for `/.well-known/did-configuration`. This setup would indeed require better documentation..
3. How would the custom headers work? Because it will be the Wallets that interact with those endpoints so how will they know what headers to use? Or are these different headers from http headers?

nanderstabel: I just noticed that in the current OID4VCI working draft (16 <-- not DIIPv4) that they've added the restriction to the `/.well-known/openid-credential-issuer` that it needs to be appended directly to the domain: https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#appendix-J-3.9

>Credential Issuers publishing metadata MUST make a JSON document available at the path formed inserting the string /.well-known/openid-credential-issuer into the Credential Issuer Identifier between the host component and the path component, if any.
>
>For example, the metadata for the Credential Issuer Identifier https://issuer.example.com/tenant would be retrieved from https://issuer.example.com/.well-known/openid-credential-issuer/tenant. The metadata for the Credential Issuer Identifier https://tenant.issuer.example.com would be retrieved from https://tenant.issuer.example.com/.well-known/openid-credential-issuer.

This is not directly relevant to us now since it is not part of DIIP yet, but it is another reason to put certain restrictions to the Public URL.. Perhaps we should indeed block base paths altogether and suggest `https://tenant.issuer.example.com` over `https://issuer.example.com/tenant`?

daniel-mader: Tracked in #196

daniel-mader (approved):

PR #193, opened 2025-05-23
feat: add support for multiple credential configurations

nanderstabel:
# Description of change
* The `credential_configurations` config variable is now replaced with the new optional `credential_configuration_file` config variable which can be used to point to a json file which contains Credential Configurations. This file can contain multiple Credential Configurations which will be added to the Credential Issuer Metadata during initialization of the Issuance State. Configurations that are provisioned this way can not be updated/removed during runtime. 
* New `POST /v0/credential-configurations` endpoint where Credential Configurations can be pushed through the API during runtime.

## BREAKING CHANGE
BREAKING CHANGE:

- When utilizing the JIT Credential flow, the `/v0/offers` endpoint now **requires** the `credentialConfigurationIds` field which should contain all the IDs of the Credential Configurations corresponding to the JIT Credential(s) to be issued.
- Schema updates to the `Offer` and `ServerConfig` aggregates now prevent UniCore from loading state from event stores created with previous versions. Since data migrations are not supported yet, a fresh database is required.

**Blocked by**
- https://github.com/impierce/ssi-agent/pull/186


agent_identity/src/document/aggregate.rs
```diff
@@ -534,23 +534,30 @@ pub mod document_tests {
         document_id: String,
         did_method: SupportedDidMethod,
         document: CoreDocument,
-        document_with_verification_method: CoreDocument,
+        document_with_es256_verification_method: CoreDocument,
+        document_with_both_verification_methods: CoreDocument,
```
daniel-mader: I suggest renaming this to `multiple_verification_methods` (and possible the one that only contains es256 to `single_verification_method`).

agent_identity/src/document/aggregate.rs
```diff
@@ -685,17 +717,39 @@ pub mod test_utils {
     }
 
     #[fixture]
-    pub fn document_with_verification_method(
+    pub fn both_verification_methods(
```
daniel-mader, line 720: See comment above

config/credential_configurations.json
daniel-mader, line 1: If this file is not present, the application panicks. This introduces a dependency on an external file, which we should avoid. It would be cleaner to remove the `config` folder and default to an empty credential configurations array and "soft fail" during runtime.

agent_issuance/src/state.rs
```diff
@@ -80,14 +86,173 @@ impl Clone for Queries {
...
+}
+
+pub async fn update_credential_configurations(state: &IssuanceState) -> anyhow::Result<()> {
+    let provisioned_credential_configurations: Vec<CredentialConfiguration> = config()
+        .credential_configuration_file
+        .as_ref()
+        .map(|file| {
+            debug!("Path to credential configuration file: {}", file.as_path().display());
+
+            let file = std::fs::read(file.as_path()).expect("Failed to read credential configuration file");
```
daniel-mader, line 210: This shouldn't panick as it introduces an external dependency. Can we fallback to and empty list of provisioned credential configurations instead?

agent_shared/src/config/mod.rs
```diff
@@ -224,25 +225,10 @@ pub struct ApplicationConfiguration {
...
-                        "name": "Verifiable Credential",
-                        "locale": "en",
-                        "logo": {
-                            "uri": "https://www.impierce.com/external/impierce-logo.png",
-                            "alt_text": "Impierce Logo"
-                        }
-                    }]
-                })
-            ).unwrap()]"#
+        Some(Box::new(Path::new("./config/credential_configurations.json").to_owned()))
```
daniel-mader: Could we load a credential configuration for the `development` profile without requiring a file to be present at this path? This should only be loaded if otherwise no other credential configuration would be present.

agent_issuance/src/server_config/command.rs
```diff
@@ -10,8 +11,27 @@ pub enum ServerConfigCommand {
...
+    UpdateIssuerDisplay {
+        display: Option<Vec<serde_json::Value>>,
+    },
+    UpdateCryptographicBindingMethods {
+        cryptographic_binding_methods_supported: Vec<String>,
+    },
+    UpdateSigningAlgorithms {
+        signing_algorithms_supported: Vec<Algorithm>,
     },
     AddCredentialConfiguration {
```
daniel-mader: Do you think renaming this to `UpdateCredentialConfiguration` reflects better what it's doing, since we're technically doing an "upsert" (update or insert if not exists)?

agent_issuance/src/server_config/command.rs
```diff
@@ -10,8 +11,27 @@ pub enum ServerConfigCommand {
     InitializeServerMetadata {
         authorization_server_metadata: Box<AuthorizationServerMetadata>,
         credential_issuer_metadata: Box<CredentialIssuerMetadata>,
+        cryptographic_binding_methods_supported: Vec<String>,
+        signing_algorithms_supported: Vec<Algorithm>,
+    },
+    UpdateIssuerUrl {
+        url: url::Url,
+    },
+    UpdateIssuerDisplay {
```
daniel-mader, line 20: Is this change related to this PR?

agent_issuance/src/server_config/event.rs
```diff
@@ -14,9 +15,35 @@ pub enum ServerConfigEvent {
...
+        cryptographic_binding_methods_supported: Vec<String>,
+        credential_issuer_metadata: Box<CredentialIssuerMetadata>,
+        credential_configurations: HashMap<String, (bool, CredentialConfigurationsSupportedObject)>,
+    },
+    SigningAlgorithmsUpdated {
+        signing_algorithms_supported: Vec<Algorithm>,
+        credential_issuer_metadata: Box<CredentialIssuerMetadata>,
+        credential_configurations: HashMap<String, (bool, CredentialConfigurationsSupportedObject)>,
     },
     CredentialConfigurationAdded {
```
daniel-mader: Maybe this needs to be changed to `CredentialConfigurationUpdated` if there is some sort of a destructive action involved (such as overwriting the previous credential configuration).

agent_issuance/src/server_config/aggregate.rs
```diff
@@ -48,65 +84,150 @@ impl Aggregate for ServerConfig {
...
-
-                let proof_types_supported = HashMap::from_iter([(
-                    ProofType::Jwt,
-                    KeyProofMetadata {
-                        proof_signing_alg_values_supported: signing_algorithms_supported.clone(),
-                    },
-                )]);
+                let proof_types_supported = into_proof_types_supported(&self.signing_algorithms_supported);
+
+                // panic!("Signing algorithms: {:?}", self.signing_algorithms_supported);
```
daniel-mader: Can this comment be removed?

daniel-mader:
Please don't forget to update the Postman collection to include an example for `/v0/credential-configurations`.

agent_issuance/src/server_config/command.rs
nanderstabel: Yes it is basically a side-effect of making the Credential Configuration objects update-able during runtime. 

Prior to this PR we simply rebuilt the `ServerConfig` aggregate anew based on the configuration. But now we need to take into account that Credential Configurations can be added during runtime, so we can not simply regenerate the `ServerConfig` purely from configuration anymore. And that means that during initialization we need to check whether any URL, display, etc.. variables need to be updated in the aggregate.

daniel-mader (approved):

PR #199, opened 2025-07-08
feat: change public `display` data during runtime

nanderstabel:
# Description of change
This pull request introduces a new Profile aggregate within the Identity Bounded Context. The purpose of this change is to create a single, centralized source of truth for the agent's display information, such as its display name and logo

Key changes include:

* New `Profile` Aggregate: A new CQRS aggregate has been added to `agent_identity` to manage profile data.
* API Endpoints: New REST endpoints have been added to manage the profile:
  * `PUT /api/v0/profiles`: Creates or replaces the singleton profile.
  * `GET /api/v0/profiles/{profile_id}`: Retrieves the current profile.
* Configuration Conflict Handling: A core feature of this change is preventing runtime modifications to a profile that was provisioned via a static configuration file. Attempting to do so will now correctly return a `409 Conflict` error with a descriptive Problem Details response.
* Temporary State Propagation: As a temporary solution to ensure other bounded contexts (Issuance and Verification) reflect the updated profile data, this PR introduces a mechanism where the global configuration is updated after a profile change. This is a known anti-pattern and is clearly marked with `TODO` comments for replacement with a proper event-driven architecture in the future.
* Documentation: New API documentation for the `409 Conflict` error has been added, and the Postman collection has been updated with the new endpoints.

n/a

* The new `/profiles` endpoints were tested manually using the updated Postman collection.


agent_api_rest/src/identity/profiles/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+    extract::State,
+    response::{IntoResponse, Response},
+    Json,
+};
+use http_api_problem::ApiError;
+use hyper::StatusCode;
+use serde::{Deserialize, Serialize};
+
+#[derive(Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
```
daniel-mader, line 20: The endpoint expects a `displayName` for PATCH, but returns a `display_name` on GET. I think we should serialize consistently (using default snake_case).

agent_api_rest/src/identity/profiles/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+        command_handler(&profile_id, &state.command.profile, command).await?;
+    }
+
+    query_profile(&state).await.map_err(IntoApiErrorExt::into_api_error)?;
+
+    Ok(StatusCode::OK.into_response())
+}
+
+#[axum_macros::debug_handler]
+pub(crate) async fn get_profile(State(state): State<IdentityState>) -> Result<Response, ApiError> {
```
daniel-mader, line 84: `id` should be omitted

agent_identity/src/profile/README.md
```diff
@@ -0,0 +1,6 @@
+# Profile
+
+This aggregate holds everything related to a profile:
+- display_name
+- logo
+- source
```
daniel-mader, line 6: `- country` is missing

agent_api_rest/src/identity/profiles/mod.rs
daniel-mader, line 20: Should we `#[serde(deny_unknown_fields)]` to throw an error when unknown keys are supplied during PATCH? Do you think this adds more explicitness or is it not necessary?
Reason I am asking this, is because I am currently getting a clean `200 OK` with no further response body when supplying `display_name` instead of `displayName`, so I thought the update was successful.

agent_api_rest/src/issuance/credential_issuer/well_known/openid_credential_issuer.rs
```diff
@@ -3,20 +3,28 @@ use agent_issuance::{
...
 #[axum_macros::debug_handler]
 pub(crate) async fn openid_credential_issuer(State(state): State<IssuanceState>) -> Result<Response, ApiError> {
     match query_handler(SERVER_CONFIG_ID, &state.query.server_config).await? {
         Some(ServerConfigView {
-            credential_issuer_metadata,
+            mut credential_issuer_metadata,
             ..
-        }) => Ok((StatusCode::OK, Json(credential_issuer_metadata)).into_response()),
+        }) => {
+            // TODO: remove this once the Identity BC is the single source of truth for display data.
```
daniel-mader: What is a "BC"?

agent_api_rest/postman/ssi-agent.postman_collection.json
```diff
@@ -1778,6 +1778,78 @@
...
+										"type": "text/javascript",
+										"packages": {}
+									}
+								}
+							],
+							"request": {
+								"method": "GET",
+								"header": [],
+								"url": {
+									"raw": "{{HOST}}/v0/profiles",
```
daniel-mader: This should be `/v0/profile`.

agent_identity/src/profile/error.rs
```diff
@@ -0,0 +1,7 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum ProfileError {
+    #[error("The resource cannot be modified at runtime because it was provisioned by a static configuration file.")]
```
daniel-mader: Are we sure it is always a "file" or could is also be supplied via an env variable?

agent_identity/src/state.rs
```diff
@@ -23,7 +28,10 @@ use itertools::iproduct;
 use jsonwebtoken::Algorithm;
 use std::collections::HashMap;
 use std::sync::Arc;
-use tracing::info;
+use tracing::{info, warn};
+
+/// The unique identifier for the server configuration.
+pub const PROFILE_ID: &str = "PROFILE-001";
```
daniel-mader, line 37: Can we mark/comment this as "internally used only" somehow, so it's clear it's not reflected to the outside?

agent_api_rest/src/issuance/credential_issuer/well_known/openid_credential_issuer.rs
daniel-mader: Bounded context?

nanderstabel: Bounded Context

agent_identity/src/profile/error.rs
nanderstabel: :smile:

agent_api_rest/src/identity/profiles/mod.rs
nanderstabel: The real problem isn't about whether we ignore unknown fields (which is standard anyway), it's the inconsistency in the API.

Going with 100% snake_case would be the path of least resistance since it's the Rust default. But you're right, camelCase is probably friendlier for our API users in the long run. The main drawback is that it's more error-prone for us; we'd have to be diligent about testing and remember to handle external types, which won't get converted automatically. It just adds a bit more overhead.

So imo best would be: keep ignoring unknown fields + make sure to use camelCase consistently (+ have automatically generated API reference? + have automatically generated wasm/typescript bindings?)

agent_api_rest/postman/ssi-agent.postman_collection.json
daniel-mader: The name of the request is still called `"Get Profile by ID"` which is outdated too, I believe.

daniel-mader (approved):

PR #204, opened 2025-07-28
feat: add `credentialStatus`, serve `Status List Token`

Oran-Dan:
# Description of change
IETF OAuth Token Status List spec has been implemented for the issuer side.
2 new endpoints:
- `/ietf-oauth-token-status-list/{path}` to fetch a Token Status List (TSL) referenced by a VC
- `/credentials/{credential_id}` already existing endpoint has now an added `patch` functionality to update the CredentialStatus

I've added unit tests in the files where i've added my code.
Existing tests had to be updated.


agent_issuance/Cargo.toml
```diff
@@ -28,6 +29,7 @@ serde_json_path_to_error.workspace = true
 thiserror.workspace = true
 tracing.workspace = true
 url.workspace = true
+rand.workspace = true
```
nanderstabel: Not alphabetical

agent_issuance/src/state.rs
```diff
@@ -95,6 +97,8 @@ pub async fn initialize(state: &IssuanceState) -> anyhow::Result<()> {
     update_signing_algorithms(state).await?;
     update_credential_configurations(state).await?;
 
+    // I think the signer needs to be initialized here, but it is not clear how to do that.
+
```
nanderstabel: Can this be removed?

agent_issuance/src/credential/event.rs
```diff
@@ -30,6 +33,10 @@ pub enum CredentialEvent {
         credential_id: String,
         notification: NotificationRequest,
     },
+    CredentialStatusSet {
```
nanderstabel: The word "Set" is correct, but also a bit ambiguous (because it is also a noun) so we prefer to use "Updated" --> `CredentialStatusUpdated`

agent_issuance/src/credential/command.rs
```diff
@@ -29,4 +38,8 @@ pub enum CredentialCommand {
         credential_id: String,
         notification: NotificationRequest,
     },
+    SetCredentialStatus {
```
nanderstabel: See https://github.com/impierce/ssi-agent/pull/204/files#r2245301950

agent_issuance/src/credential/aggregate.rs
```diff
@@ -577,9 +668,9 @@ pub mod test_utils {
...
...SzljZjVXNEc4Z2NaOXh1d0p2RzFlN3dOazhLQ2d0I3o2TWtnRTg0TkNNcE1lQXg5aks5Y2Y1VzRHOGdjWjl4dXdKdkcxZTd3Tms4S0NndCJ9.eyJpc3MiOiJkaWQ6a2V5Ono2TWtnRTg0TkNNcE1lQXg5aks5Y2Y1VzRHOGdjWjl4dXdKdkcxZTd3Tms4S0NndCIsInN1YiI6ImRpZDprZXk6ejZNa2dFODROQ01wTWVBeDlqSzljZjVXNEc4Z2NaOXh1d0p2RzFlN3dOazhLQ2d0IiwibmJmIjoxMjYyMzA0MDAwLCJpYXQiOjEyNjIzMDQwMDAsInZjIjp7IkBjb250ZXh0IjpbImh0dHBzOi8vd3d3LnczLm9yZy8yMDE4L2NyZWRlbnRpYWxzL3YxIl0sInR5cGUiOlsiVmVyaWZpYWJsZUNyZWRlbnRpYWwiXSwiY3JlZGVudGlhbFN1YmplY3QiOnsiaWQiOiJkaWQ6a2V5Ono2TWtnRTg0TkNNcE1lQXg5aks5Y2Y1VzRHOGdjWjl4dXdKdkcxZTd3Tms4S0NndCIsImZpcnN0X25hbWUiOiJGZXJyaXMiLCJsYXN0X25hbWUiOiJSdXN0YWNlYW4iLCJkZWdyZWUiOnsidHlwZSI6Ik1hc3RlckRlZ3JlZSIsIm5hbWUiOiJNYXN0ZXIgb2YgT2NlYW5vZ3JhcGh5In19LCJpc3N1ZXIiOiJkaWQ6a2V5Ono2TWtnRTg0TkNNcE1lQXg5aks5Y2Y1VzRHOGdjWjl4dXdKdkcxZTd3Tms4S0NndCIsImlzc3VhbmNlRGF0ZSI6IjIwMTAtMDEtMDFUMDA6MDA6MDBaIiwiY3JlZGVudGlhbFN0YXR1cyI6eyJpZCI6Imh0dHBzOi8vbXktZG9tYWluLmV4YW1wbGUub3JnL2lldGYtb2F1dGgtdG9rZW4tc3RhdHVzLWxpc3QvMSIsInR5cGUiOiJzdGF0dXNsaXN0K2p3dCIsInVyaSI6Imh0dHBzOi8vbXktZG9tYWluLmV4YW1wbGUub3JnL2lldGYtb2F1dGgtdG9rZW4tc3RhdHVzLWxpc3QvMSIsImlkeCI6MH19fQ.6-i-8_wL-TcqzW6s73lMSgFW4R5eLGT5vDSFw-RuJi_oz2HsKPcbkzL8fhqE49rpyeOHHiJDp6RAw2uKkTdEAA";
```
nanderstabel: I guess only now I (partially) understand what [this issue](https://github.com/FIDEScommunity/DIIP/issues/60) is about, because I expected there to be a `status` object in the root like this one:
```json
    ...
    ,
   {
     "status": {
       "status_list": {
         "idx": 0,
         "uri": "https://example.com/statuslists/1"
       }
     }
   }
```


 in this VC (`W3C_VC_VERIFIABLE_CREDENTIAL_JWT` decoded):
```json
{
  "iss": "did:jwk:eyJhbGciOiJFUzI1NiIsImNydiI6IlAtMjU2Iiwia2lkIjoiOEJkS25MLTI0LVM0eXd6OEhXa044akdKOTF3dl9ZcUtCYmVzZFJfenB1byIsImt0eSI6IkVDIiwieCI6Im9udHc1dGYyblJzcGFkaXJiWTBLTU12bGlfZWJwSDJvRld4WnlFalBpb3MiLCJ5Ijoiak9nVkpzV3pjTWNPX1U1RUxWQ1dPbHp4bE5obnFoby1NdmNoRmhGeXZsayJ9",
  "sub": "did:key:z6MkuiRKq1fKrzAXeSNiGwrpJPPugY8AxJYA5cpCvZCYBD7B",
  "nbf": 1753967429,
  "iat": 1753967429,
  "vc": {
    "@context": [
      "https://www.w3.org/2018/credentials/v1"
    ],
    "type": [
      "VerifiableCredential"
    ],
    "credentialSubject": {
      "id": "did:key:z6MkuiRKq1fKrzAXeSNiGwrpJPPugY8AxJYA5cpCvZCYBD7B",
      "dob": "1982-01-01",
      "first_name": "Ferris",
      "last_name": "Crabman"
    },
    "issuer": "did:jwk:eyJhbGciOiJFUzI1NiIsImNydiI6IlAtMjU2Iiwia2lkIjoiOEJkS25MLTI0LVM0eXd6OEhXa044akdKOTF3dl9ZcUtCYmVzZFJfenB1byIsImt0eSI6IkVDIiwieCI6Im9udHc1dGYyblJzcGFkaXJiWTBLTU12bGlfZWJwSDJvRld4WnlFalBpb3MiLCJ5Ijoiak9nVkpzV3pjTWNPX1U1RUxWQ1dPbHp4bE5obnFoby1NdmNoRmhGeXZsayJ9",
    "issuanceDate": "2025-07-31T13:10:29Z",
    "credentialStatus": {
      "id": "http://localhost:3033/ietf-oauth-token-status-list/0",
      "type": "statuslist+jwt",
      "idx": 307,
      "uri": "http://localhost:3033/ietf-oauth-token-status-list/0"
    }
  }
}
```

as described in 6.2.  Referenced Token in JOSE.

In addition, in 8.3.  Validation Rules it is stated that (after the 'usual' JWT validation):
>    1.  Check for the existence of a status claim, check for the
       existence of a status_list claim within the status claim and
       validate that the content of status_list adheres to the rules
       defined in Section 6.2 for JOSE-based Referenced Tokens and
       Section 6.3 for COSE-based Referenced Tokens.  Other formats of
       Referenced Tokens may define other encoding of the URI and index.

but currently our VCs don't have a `status` claim and therefore also no nested `status_list` claim.

So is that the intended implementation, because it seems to me like what we currently have is not conform the Token Status List spec? Or does _"Other formats of Referenced Tokens may define other encoding of the URI and index"_ mean that using the "credentialStatus" property in the case of W3C VC DM is permitted? And if so, then why is `"statuslist+jwt"` is the correct value for `"type"` and not for example `"TokenStatusList"` which at least on face value seems closer related to [the description of type in W3C VC DM](https://www.w3.org/TR/vc-data-model-2.0/#types).

Perhaps some more context as to why I expect there to be a `"status"` field is the nuance that UniCore does not support W3C VC DM in its 'pure' form, but rather indirectly through the [`jwt_vc_json` Credential Format Identifier](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#appendix-A.1.1.1-1) (which is a JWT). So to me it makes the most sense that the `jwt_vc_json`'s (as well as the [`dc+sd-jwt`'s](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#appendix-A.1.1.1-1)) that we produce all have this `status` claim in the root level since this will be 100% conform to the Token Status List spec. I definitely agree with @muisit that the DIIPv4 spec is unclear about this.

Is there some specification that I missed or misunderstood? Or has there been some consensus about this issue already during the DIIP meetings that I am not aware of?

TL;DR: I don't understand why there's no `status` claim :joy:

agent_api_rest/Cargo.toml
```diff
@@ -46,6 +49,7 @@ agent_secret_manager = { path = "../agent_secret_manager", features = ["test_uti
 agent_shared = { path = "../agent_shared", features = ["test_utils"] }
 agent_store = { path = "../agent_store" }
 agent_verification = { path = "../agent_verification", features = ["test_utils"] }
+agent_api_rest = { path = ".", features = ["test_utils"] }
```
nanderstabel: Not alphabetical

agent_api_rest/src/issuance/credentials.rs
```diff
@@ -149,6 +197,38 @@ pub(crate) async fn all_credentials(State(state): State<IssuanceState>) -> Resul
     Ok((StatusCode::OK, Json(all_credentials)).into_response())
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub struct PatchCredentialEndpointRequest {
+    pub credential_status: StatusType,
+}
```
nanderstabel: We are not consistently doing this right now (at all), but we need to make sure that for the UniCore-specific API everything is camelCase (through `#[serde(rename_all = "camelCase")]`) (see: https://github.com/impierce/ssi-agent/pull/199#discussion_r2239377454)

nanderstabel (changes requested):

agent_issuance/Cargo.toml
Oran-Dan: fixed

agent_issuance/src/state.rs
Oran-Dan: fixed

agent_issuance/src/credential/event.rs
Oran-Dan: renamed

Oran-Dan: i've also renamed the command to be coherent

agent_issuance/src/credential/command.rs
Oran-Dan: renamed

agent_api_rest/Cargo.toml
Oran-Dan: fixed

agent_api_rest/src/issuance/credentials.rs
Oran-Dan: fixed

agent_issuance/Cargo.toml
```diff
@@ -17,10 +17,12 @@ derivative.workspace = true
 identity_core.workspace = true
 identity_credential.workspace = true
 jsonwebtoken.workspace = true
+oauth_tsl.workspace = true
 oid4vci.workspace = true
 oid4vc-core.workspace = true
 strum.workspace = true
 strum_macros.workspace = true
```
nanderstabel: Looks like these two were already out of order, can you perhaps fix that?

agent_api_rest/src/issuance/credentials.rs
```diff
@@ -13,9 +18,14 @@ use axum::{
 };
 use http_api_problem::ApiError;
 use hyper::header;
+use oauth_tsl::status_list::StatusType;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
+pub const STATUSTYPESIZE: u8 = 2; // Amount of bits per status
+pub const STATUSLISTSIZE: usize = 2048; // Amount of bytes in the status list. Equates to 8192 statuses for STATUSTYPESIZE = 2.
+pub const TESTINDEX: usize = 123;
```
nanderstabel, line 22: The only purpose of this constant is to be used during testing within this crate so it should not be included in non-test builds. 

You can just move it into the `pub mod tests` scope below and replace both occurrences of `feature = "test_utils"` with just `test`.

agent_api_rest/src/issuance/credentials.rs
```diff
@@ -13,9 +18,14 @@ use axum::{
 };
 use http_api_problem::ApiError;
 use hyper::header;
+use oauth_tsl::status_list::StatusType;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
+pub const STATUSTYPESIZE: u8 = 2; // Amount of bits per status
```
nanderstabel: imo the name for this constant is a bit ambiguous so consider renaming it to something more explicit like `BITS_PER_STATUS`?

agent_api_rest/src/issuance/credentials.rs
```diff
@@ -13,9 +18,14 @@ use axum::{
...
+pub const STATUSLISTSIZE: usize = 2048; // Amount of bytes in the status list. Equates to 8192 statuses for STATUSTYPESIZE = 2.
```
nanderstabel: This could maybe also be a bit more explicit? In general imo it is a bit confusing that the SIZE in the first constant refers to (amount of) bits, but in this second const it refers to (amount of) bytes.

Something like: `STATUS_LIST_SIZE_IN_BYTES` or `STATUS_LIST_BYTES`?

agent_api_rest/src/issuance/credentials.rs
```diff
@@ -94,11 +104,49 @@ pub(crate) async fn credentials(
...
+
+        let random_index;
+        #[cfg(not(feature = "test_utils"))]
+        {
+            use rand::Rng;
+
+            let mut rng = rand::rng();
+            loop {
+                let candidate = rng
+                    .random_range((status_list_number * STATUSLISTSIZE)..((status_list_number + 1) * STATUSLISTSIZE));
```
nanderstabel: `(status_list_number * STATUSLISTSIZE)` and `((status_list_number + 1) * STATUSLISTSIZE)` don't need to be calculated over and over again right? Can they be be calculated prior to the loop and then reused?

agent_api_rest/src/issuance/mod.rs
```diff
@@ -26,7 +30,10 @@ pub fn router(issuance_state: IssuanceState) -> Router {
             API_VERSION,
             Router::new()
                 .route("/credentials", post(credentials).get(all_credentials))
-                .route("/credentials/{credential_id}", get(credentials::credential))
+                .route(
+                    "/credentials/{credential_id}",
+                    get(credentials::credential).patch(patch_credential),
```
nanderstabel, line 36: For both endpoints that you added can you please also add them to the Postman collection? :)

agent_api_rest/src/issuance/mod.rs
```diff
@@ -1,18 +1,22 @@
 pub mod credential_configurations;
 pub mod credential_issuer;
 pub mod credentials;
-pub mod offers;
-
 pub mod error;
+pub mod offers;
```
nanderstabel: nit: can you add this back to line 4 with an empty line in line 5? This way the list of modules that all contain endpoint handlers stay together while the error module stays separate

agent_issuance/src/credential/aggregate.rs
```diff
@@ -68,6 +70,13 @@ mod never_as_str {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, Default, Derivative)]
+#[derivative(PartialEq)]
```
nanderstabel: `#[derivative(PartialEq)]` allows you to add certain attributes to the struct fields (for example to skip a field when comparing two instances). Since we don't make use of those attributes this can be removed and `Derivative` can be replaced with `PartialEq`. You can apply the same changes for `Credential` which also doesn't use those attributes (anymore).

agent_issuance/src/credential/aggregate.rs
```diff
@@ -225,6 +262,11 @@ impl Aggregate for Credential {
                                 )
                                 .map_err(|e| InvalidCredentialSubjectError(e.to_string()))?;
 
+                                let builder_credential_status = types_ob_v3::prelude::CredentialStatus {
+                                    id: status_list_url.clone().into(),
```
nanderstabel: nit: to be a bit more explicit --> `status_list_url.to_string()`

agent_api_rest/src/issuance/credential_issuer/token_status_list.rs
```diff
@@ -0,0 +1,211 @@
...
+        let mut indices = used_indices.clone();
+
+        let mut rng = rand::rng();
+        while indices.len() < amount_indices {
+            let random_index = rng.random_range(lower_bound..upper_bound);
+            if !indices
+                .iter()
+                .any(|credential_status| credential_status.index == random_index)
+            {
+                let status_type = rng.random_range(0..2);
```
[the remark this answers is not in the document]
nanderstabel: I think Copilot is correct about this right?

agent_api_rest/src/issuance/credential_issuer/token_status_list.rs
```diff
@@ -0,0 +1,211 @@
...
+                Some(c.credential_status.clone())
+            } else {
+                None
+            }
+        })
+        .collect();
+
+    // This block ensures that the remaining empty 30% of a status list is filled with random values.
+    // This block works in tandem with the part of `fn patch_credential` which only fills 70% of a status list.
+    used_indices = task::spawn_blocking(move || -> Result<Vec<CredentialStatus>, PublicError> {
```
nanderstabel: I don't think it is necessary to use `task::spawn_blocking` here

agent_api_rest/src/issuance/credential_issuer/token_status_list.rs
```diff
@@ -0,0 +1,211 @@
...
+        StatusCode::OK,
+        [
+            (
+                header::CONTENT_TYPE,
+                StatusListTokenResponseType::Jwt.to_string().as_str(),
+            ),
+            (header::CONTENT_ENCODING, "gzip"),
+        ],
+        compressed_jwt_token,
+    )
```
nanderstabel, line 123: I'm not really familiar with this `gzip` content encoding but currently it causes this line to panic in our middleware: `info!("Response Body: {}", std::str::from_utf8(chunk).unwrap());`

So to fix that you need to get rid of that unwrap (which was already very bad in the first place) by only logging the message if `std::str::from_utf8(chunk)` returns `Ok`

nanderstabel (changes requested):

agent_issuance/src/credential/aggregate.rs
Oran-Dan: fixed

agent_api_rest/src/issuance/credentials.rs
Oran-Dan: fixed

Oran-Dan: actually after runnning tests, i found out at least agent_holder uses this too, I'm assuming all tests in other folders which also create a signed or unsigned credential use this. Wasn't immediately clear to me either but i'll have to keep it in

Oran-Dan: I've moved TESTINDEX to the only file where it is used, that's all

agent_api_rest/src/issuance/credentials.rs
Oran-Dan: fixed

agent_api_rest/src/issuance/credentials.rs
Oran-Dan: fixed

agent_api_rest/src/issuance/credentials.rs
Oran-Dan: fixed

agent_issuance/src/credential/aggregate.rs
Oran-Dan: fixed

agent_issuance/src/credential/aggregate.rs
Oran-Dan: fixed

agent_api_rest/src/issuance/credential_issuer/token_status_list.rs
Oran-Dan: removed

agent_api_rest/src/issuance/credentials.rs
nanderstabel: Ah ok understood, then still I'm going to be annoying an ask you to add to `"test_utils"` attribute to it :see_no_evil: 
```rust
#[cfg(feature = "test_utils")]
pub const TESTINDEX: usize = 123;
```

Oran-Dan: fixed

agent_api_rest/src/issuance/mod.rs
Oran-Dan: fixed

agent_api_rest/src/issuance/mod.rs
Oran-Dan: Done

agent_api_rest/postman/ssi-agent.postman_collection.json
```diff
@@ -42,7 +43,7 @@
...
+							"raw": "{\n    \"offerId\":\"{{OFFER_ID}}\",\n    \"credentialConfigurationId\": \"001\",\n    \"credential\": {\n        \"credentialSubject\": {\n            \"type\": [ \"AchievementSubject\" ],\n            \"achievement\": {\n                \"id\": \"https://example.com/achievements/21st-century-skills/teamwork\",\n                \"type\": \"Achievement\",\n                \"criteria\": {\n                    \"narrative\": \"Team members are nominated for this badge by their peers and recognized upon review by Example Corp management.\"\n                },\n                \"description\": \"This badge recognizes the development of the capacity to collaborate within a group environment.\",\n                \"name\": \"Teamwork\"\n            }\n        }\n    },\n    \"expiresAt\": \"2028-04-12T09:15:23Z\"\n}",
```
nanderstabel: Let's leave the id for this one as `openbadge_credential`. It's ok that not all requests work directly 'out of the box' for the development profile. This request example is explicitly for OpenBadge credentials but the development default for the credential configuration with id: `"001"` is 'just' a regular W3C VC. That without any other configuration this request will return a 4xx status code is not a bad thing per se, it just means that it demonstrates that a specific Problem Details error response is returned unless you configure a new credential configuration with id: `openbadge_credential`

agent_api_rest/postman/ssi-agent.postman_collection.json
```diff
@@ -224,6 +225,69 @@
...
+							"raw": "{{HOST}}{{CREDENTIAL_LOCATION}}",
+							"host": [
+								"{{HOST}}{{CREDENTIAL_LOCATION}}"
+							]
+						}
+					},
+					"response": []
+				},
+				{
+					"name": "Status List Token by ID",
```
nanderstabel: nit: `... by ID"` is slightly misleading since the Status List doesn't really have an ID (right now in the collection we use that only for querying instances of Aggregates)

so just naming it `Status List Token` would suffice imo

agent_api_rest/src/issuance/error.rs
```diff
@@ -41,6 +41,16 @@ impl IntoApiErrorExt for CredentialError {
...
+            InvalidCredentialStatus => ApiError::builder(StatusCode::INTERNAL_SERVER_ERROR)
+                .title("Invalid Credential Status URL")
+                .type_url(type_url("issuance#invalid-credential-status-url"))
+                .source(self)
+                .finish(),
+            BuildVcJwtError(_) => ApiError::builder(StatusCode::INTERNAL_SERVER_ERROR)
+                .title("Failed to Create VC JWT")
+                .type_url(type_url("issuance#build-vc-jwt-error"))
+                .source(self)
+                .finish(),
```
nanderstabel: The purpose of `ApiError` is to provide the user extra details or context as to why a specific error occurred and more importantly how to resolve the problem. The `type_url` value will refer to a specific paragraph in our Problem Details documentation that contains that extra context.

You can find the source of our problem details in `./docs/problem-details/*.md`.

However, I don't think for these two errors that you defined you need to write any additional Problem Detail in the docs. These two errors are not a result of any misconfiguration, or misuse of the API bu the user of UniCore. These errors _should_ never occur, and if they do occur then that means it is the result of a bug and therefore the only way to resolve the errors is for us to fix the bug. For those kind of errors we have the `Unexpected Error` problem detail (see `./docs/problem-details/unexpected.md`) which means you can simply replace the `ApiError`'s that you added here with this:
```rust
ApiError::builder(StatusCode::INTERNAL_SERVER_ERROR)
                .title("Unexpected Error")
                .type_url(format!(
                    "{DOCUMENTATION_URL}problem-details/unexpected#unexpected-error"
                ))
                .source(self)
                .finish()
```

agent_shared/src/config/mod.rs
```diff
@@ -24,8 +24,10 @@ use crate::{error::SharedError, profile::ApplicationProfile};
 // Re-export
 pub use provisioned::load_provisioned_config;
 
-static STRONGHOLD_PATH: &str = "./stronghold.dat";
+pub const BITS_PER_STATUS: u8 = 2; // Amount of bits per status
+pub const STATUS_LIST_BYTES_AMOUNT: usize = 2048; // Amount of bytes in the status list. Equates to 8192 statuses for BITS_PER_STATUS = 2.
 
+static STRONGHOLD_PATH: &str = "./stronghold.dat";
```
nanderstabel, line 30: nit: can you add back the empty line under this line?

nanderstabel (changes requested):

agent_api_rest/postman/ssi-agent.postman_collection.json
Oran-Dan: oke done

agent_api_rest/postman/ssi-agent.postman_collection.json
Oran-Dan: fixed

agent_shared/src/config/mod.rs
Oran-Dan: fixed

nanderstabel (approved):

PR #213, opened 2025-08-24
feat: add transaction code support to OID4VCI flow

coplat:
# Description of change
Adds the `tx_code` (transaction code) parameter to the Credential Offer for pre-authorized code flows, as defined in the OID4VCI [specification](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#token-request). 

Necessary additions were also made to the token request within the offer aggregate where the creation of a unique transaction code is initiated when `tx_code` is a required. This flow is initialized when when/if constraints are provided in the Credential Offer (`TxCodeConstraints`). A Token Response is then created if the code in the Token Request is a match to the expected `tx_code`. 

Added simple tests which validate that the generated tx_codes are as specified in the constraints provided within the Credential Offer. 


agent_issuance/src/offer/command.rs
```diff
@@ -25,7 +28,6 @@ pub enum OfferCommand {
         offer_id: String,
         target_url: Url,
     },
-
```
nanderstabel, line 28: This empty line here indicates that the  `// OpenID4VCI Pre-Authorized Code Flow` comment is scoped to all Command variants beneath it and not just to `CreateTokenResponse`, so it should not be removed.

agent_issuance/src/utils/generate_tx_code.rs
```diff
@@ -0,0 +1,58 @@
+use oid4vci::credential_offer::{InputMode, TxCodeConstraints};
+use rand::Rng;
+// This function will generate a transaction code based on parameters defined in the TxCode struct within a credential offer.
+// These parameters include input mode, length, and description. It needs to be then sent to the user via email.
+
+pub fn generate_tx_code(tx_code: &TxCodeConstraints) -> String {
```
nanderstabel, line 7: There should be an empty line under `use rand::Rng;` but there should not be an empty line between the comments and `pub fn generate_tx_code(tx_code: &TxCodeConstraints) -> String {`. I would also suggest to use doc-comments here (`///`).

agent_api_rest/src/issuance/offers/mod.rs
```diff
@@ -62,6 +63,8 @@ pub(crate) async fn offers(
         let command = OfferCommand::CreateCredentialOffer {
             offer_id: offer_id.clone(),
             credential_configuration_ids,
+            grant_types: vec![GrantType::PreAuthorizedCode],
+            tx_code_constraints: None,
```
nanderstabel: Since here as well as in `agent_api_rest/src/issuance/credentials.rs` `tx_code_constraints: None` is used there is currently actually no we for UniCore 'consumers' to make use of the TX Code functionality that you've implemented in the domain layer. 

We previously talked about enabling tx code through the Request body to `/v0/credentials` and/or `/v0/offers`, but after talking with @JelleMillenaar it actually made more sense to me to configure this through the Credential Configuration.

agent_api_rest/src/issuance/error.rs
```diff
@@ -99,6 +99,9 @@ impl IntoApiErrorExt for OfferError {
             MissingProofError => ApiError::new(StatusCode::INTERNAL_SERVER_ERROR),
             InvalidProofError(_) => ApiError::new(StatusCode::INTERNAL_SERVER_ERROR),
             MissingProofIssuerError => ApiError::new(StatusCode::INTERNAL_SERVER_ERROR),
+            MissingTxCodeError => ApiError::new(StatusCode::BAD_REQUEST),
+            InvalidTxCodeError => ApiError::new(StatusCode::BAD_REQUEST),
+            InvalidPreAuthorizedCodeError => ApiError::new(StatusCode::INTERNAL_SERVER_ERROR),
```
nanderstabel: All these errors can only occur through the `CreateTokenResponse` command which in turn is only utilized from the `/auth/token` endpoint, which is a Public endpoint. `ApiError` is only relevant for non-public endpoints (meaning UniCore API endpoints) as they serve to give extra context to someone integrating with UniCore. 

All the error variants listed under `// Public API Errors` should be unreachable and therefore if somehow they are reachable due to a bug, then it should return `INTERNAL_SERVER_ERROR` (instead of `BAD_REQUEST`).

Other than that, since the errors that you've added can only occur through the `/auth/token` endpoint, they should be move up under the `// `/auth/token` endpoint` comment.

agent_issuance/src/offer/aggregate.rs
```diff
@@ -191,16 +213,49 @@ impl Aggregate for Offer {
...
+                        if let Some(CredentialOffer::CredentialOffer(params)) = &self.credential_offer {
+                            params
+                                .grants
+                                .as_ref()
+                                .and_then(|grants| grants.pre_authorized_code.as_ref())
+                                .and_then(|pre_auth| pre_auth.tx_code.as_ref())
+                                .is_some()
+                        } else {
+                            false
+                        };
```
nanderstabel: Could this be simplified to `let offer_requires_tx_code = self.tx_code.is_some();`?

agent_issuance/src/offer/aggregate.rs
```diff
@@ -191,16 +213,49 @@ impl Aggregate for Offer {
...
+                    match (offer_requires_tx_code, tx_code) {
+                        (true, None) => return Err(MissingTxCodeError),
+                        (true, Some(provided_tx_code)) => {
+                            let expected_tx_code = self.tx_code.as_ref().ok_or(MissingTxCodeError)?;
+
+                            if provided_tx_code != *expected_tx_code {
+                                return Err(InvalidTxCodeError);
+                            }
+                        }
+                        (false, _) => {}
```
nanderstabel, line 236: I think there should also be a rule for the case when the Wallet does [send a Transaction Code while it is not required](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#section-6.3-4.1).

agent_issuance/src/offer/error.rs
```diff
@@ -18,4 +18,10 @@ pub enum OfferError {
     UnsupportedTokenRequestGrantTypeError,
     #[error("Invalid `credential_offer_uri`: {0}")]
     InvalidCredentialOfferUriError(#[source] url::ParseError),
+    #[error("Transaction code is missing.")]
+    MissingTxCodeError,
+    #[error("Wrong transaction code provided.")]
+    InvalidTxCodeError,
+    #[error("Pre-Authorized Code is invalid.")]
+    InvalidPreAuthorizedCodeError,
```
nanderstabel, line 28: These are Public API errors corresponding to the [Token Error Responses](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#section-6.3), which means they need to be mapped in the `IntoPublicError` `impl` blok for `OfferError` that currently looks like this:
```rust
impl IntoPublicError for OfferError {
    fn into_public_error(self) -> PublicError {
        use OfferError::*;
        match self {
            MissingCredentialOfferError => {
                PublicError::CredentialError(OID4VCError::new(CredentialErrorResponse::InvalidCredentialRequest))
            }
            // TODO: check for missing error responses
            _ => PublicError::InternalServerError,
        }
    }
}
```

For each of these Token Error Responses you need to map it to the corresponding error variant that you have defined in our openid4vc crate:
```rust
#[derive(Debug, Serialize, Deserialize)]
#[serde(rename_all = "snake_case")]
pub enum TokenErrorResponse {
    InvalidRequest,
    InvalidClient,
    InvalidGrant,
    UnauthorizedClient,
    UnsupportedGrantType,
    InvalidScope,
}
```

(similar to how we currently doing it for this error:
```rust
MissingCredentialOfferError => {
                PublicError::CredentialError(OID4VCError::new(CredentialErrorResponse::InvalidCredentialRequest))
            }```)

File: agent_api_rest/src/issuance/offers/mod.rs (hunk 3 above)
Comment (nanderstabel) [reply agent_api_rest/src/issuance/offers/mod.rs]: I think what would be the best way of doing this is to add a field to this struct that we currently have:
```rust
#[derive(Deserialize, Serialize, Debug, Clone)]
pub struct CredentialConfiguration {
    pub credential_configuration_id: String,
    #[serde(flatten)]
    pub credential_format_with_parameters: CredentialFormats<WithParameters>,
    #[serde(default)]
    pub display: Vec<CredentialConfigurationsSupportedDisplay>,
    #[serde(default)]
    pub claims: Vec<ClaimDescription>,
}
```

Perhaps a field like this:
```rust

    pub authorization_grant: AuthorizationGrant,
}

#[derive(Deserialize, Serialize, Debug, Clone)]
pub enum AuthorizationGrant {
    PreAuthorized {
        tx_code_constraints: Option<TxCodeConstraints>,
    },
}
```

Then, in `agent_api_rest/src/issuance/credentials.rs` and `agent_api_rest/src/issuance/offers/mod.rs` you should be able to collect `tx_code_constraints` so that you can use it in the `CreateCredentialOffer` command.

nanderstabel (changes requested):

agent_api_rest/src/issuance/offers/mod.rs
coplat: Ok so to do this I had to add an additional `pub credential_authorization_grants: HashMap<String, AuthorizationGrant>` field to the ServerConfig struct as it wouldn't be possible to add the authorization_grant field to `CredentialConfigurationsSupportedObject` since that is directly from the OID4VCI spec.

agent_api_rest/src/issuance/error.rs
```diff
@@ -177,7 +181,13 @@ impl IntoPublicError for OfferError {
...
                 PublicError::CredentialError(OID4VCError::new(CredentialErrorResponse::InvalidCredentialRequest))
             }
-            // TODO: check for missing error responses
+            MissingTxCodeError => PublicError::TokenError(OID4VCError::new(TokenErrorResponse::InvalidRequest)),
+            InvalidTxCodeError => PublicError::TokenError(OID4VCError::new(TokenErrorResponse::InvalidGrant)),
+            InvalidPreAuthorizedCodeError => {
+                PublicError::TokenError(OID4VCError::new(TokenErrorResponse::InvalidGrant))
+            }
+            UnrequestedTxCodeError => PublicError::TokenError(OID4VCError::new(TokenErrorResponse::InvalidRequest)),
+            // TODO: check for maissing error responses
```
nanderstabel: typo: `maissing` --> `missing`

agent_shared/src/config/mod.rs
```diff
@@ -513,8 +516,15 @@ pub struct CredentialConfiguration {
...
     pub claims: Vec<ClaimDescription>,
+    pub authorization_grant: AuthorizationGrant,
 }
 
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq)]
+pub enum AuthorizationGrant {
+    PreAuthorized {
+        tx_code_constraints: Option<TxCodeConstraints>,
+    },
+}
```
nanderstabel, line 528: There should be an empty line below this block.

nanderstabel: I'm also rethinking the type now actually.. This is 'consumer-facing' API so we need to make sure that for UniCore-consumers this functionality is intuitive. This is a bit too confusing I think:
```json
{
  ...,
  "authorization_grant": {
      "PreAuthorized": {
          "tx_code_constraints": {
              "input_mode": "numeric",
              "length": 6,
              "description": "Transaction code to authorize the credential issuance"
          }
      }
  }
}
```

Better would be something like this IMO:
```json
{
  ...,
  "authorization": {
      "pre_authorized": true,
      "transaction_code_constraints": {
          "input_mode": "numeric",
          "length": 6,
          "description": "Transaction code to authorize the credential issuance"
      }
  }
}
```

1. `authorization_grant` --> `authorization` (grant is too technical and probably irrelevant to add here)
2. remove weirdly nested `PreAuthorized` object (which uses other casing than the rest of the date)
3. `"pre_authorized": true,` <-- this will be useful when we add Authorization Code flow (`"pre_authorized": false,`).
4. `tx_code_constraints` --> `transaction_code_constraints` better to use more specific naming here

Changing the `AuthorizationGrant` enum to this would simplify it a bit:
```rust
#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Default)]
pub struct Authorization {
    pub pre_authorized: bool,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub tx_code_constraints: Option<TxCodeConstraints>,
}

// Sensible default
impl Default for Authorization {
    fn default() -> Self {
        Self {
            pre_authorized: true,
            tx_code_constraints: None,
        }
    }
}
```

agent_issuance/src/state.rs
```diff
@@ -216,47 +216,52 @@ pub async fn update_credential_configurations(state: &IssuanceState) -> anyhow::
...
+                                        "locale": "en"
+                                    }],
+                                }
+                            ],
+                            "authorization_grant": {
+                                "PreAuthorized": {
+                                    "tx_code_constraints": null
                   }
-                ]))
+                }
```
nanderstabel: You had to add this code here because `authorization_grant` doesn't have a sensible default (which it should, see https://github.com/impierce/ssi-agent/pull/213/files#r2392404239).

After you've implemented the sensible default you should be able to revert all the changes in this JSON block

agent_issuance/src/utils/generate_tx_code.rs
```diff
@@ -0,0 +1,59 @@
...
+/// These parameters include input mode, length, and description. The generated Transaction Code (TxCode) will then be sent to the end user
+/// out-of-band, e.g. via e-mail.
+pub fn generate_tx_code(tx_code: &TxCodeConstraints) -> String {
+    let length = tx_code.length.unwrap_or(6) as usize; // Default to 6 digits if not otherwise specified
+    let input_mode = tx_code.input_mode.as_ref().unwrap_or(&InputMode::Numeric);
+    let mut rng = rand::rng();
+
+    match input_mode {
+        InputMode::Numeric => {
+            // Generate 6 random digits.
```
nanderstabel: nit: change this to: `// Generate random digits.` (since `6` only applies to the default case). Same goes for the comment in line 18

nanderstabel (changes requested):

agent_shared/src/config/mod.rs
```diff
@@ -513,6 +516,23 @@ pub struct CredentialConfiguration {
     pub display: Vec<CredentialConfigurationsSupportedDisplay>,
     #[serde(default)]
     pub claims: Vec<ClaimDescription>,
+    pub authorization: Authorization,
```
nanderstabel, line 520: Running UniCore with the development profile currently panics:
```bash
2025-10-01T11:38:58.820290Z  INFO agent_issuance::state: Using default development credential configurations.

thread 'main' panicked at /home/nander/Impierce/ssi-agent/agent_issuance/src/state.rs:260:18:
Failed to parse default development credential configurations: Error("missing field `authorization`", line: 0, column: 0)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

but that can easily be fixed by adding the `#[serde(default)]` attribute on top of `pub authorization: Authorization,` :)

agent_shared/tests/test_credential_configurations.json
```diff
@@ -14,6 +14,10 @@
           "alt_text": "Impierce Logo"
         }
       }
-    ]
-  }
-]
+    ],
+    "authorization": {
+      "pre_authorized": true,
+      "transaction_code_constraints": null
```
nanderstabel: `transaction_code_constraints` is optional now so it looks a big cleaner to just omit it altogether instead of assigning a `null` value to it.

agent_shared/tests/test_credential_configurations.json
```diff
@@ -14,6 +14,10 @@
...
+    }
+      }
```
nanderstabel: The indentation of this line should align with the `{` in line 2

agent_issuance/src/offer/aggregate.rs
```diff
@@ -243,8 +289,7 @@ impl Aggregate for Offer {
                 let credential_response = CredentialResponse {
                     credential: CredentialResponseType::Immediate {
                         credentials: vec![CredentialResponseObject {
-                            // TODO: Apply strong typing to signed credentials.
-                            credential: signed_credential.as_str().unwrap_or_default().to_string(),
```
nanderstabel, line 247: Changing these lines was not intentional I think?

agent_issuance/src/offer/aggregate.rs
```diff
@@ -258,7 +303,6 @@ impl Aggregate for Offer {
             }
         }
     }
-
```
nanderstabel, line 261: This empty line should stay here

nanderstabel (changes requested):

agent_api_rest/src/issuance/credentials.rs
```diff
@@ -149,9 +151,36 @@ pub(crate) async fn credentials(
...
+                    tx_code_constraints: Some(tx_code_constraints),
+                } => Some(tx_code_constraints.clone()),
+                Authorization {
+                    pre_authorized: true,
+                    tx_code_constraints: None,
+                } => None,
+                Authorization {
+                    pre_authorized: false, ..
+                } => None,
+            });
```
nanderstabel: We've already queried the `server_config` in line 61. If you update the codeblock starting from line 61 from:
```rust
    let credential_configuration = query_handler(SERVER_CONFIG_ID, &state.query.server_config)
        .await?
        .and_then(|server_config_view| {
            server_config_view
                .credential_issuer_metadata
                .credential_configurations_supported
                .get(&credential_configuration_id)
                .cloned()
        })
```

to 
```rust
    let (_, credential_configuration, authorization) = query_handler(SERVER_CONFIG_ID, &state.query.server_config)
        .await?
        .and_then(|server_config_view| {
            server_config_view
                .credential_configurations
                .get(&credential_configuration_id)
                .cloned()
        })
```

you will already have access to the `authorization`. You can then simplify the code block starting at line 157 to:
```rust
        let tx_code_constraints = match authorization {
            Authorization {
                pre_authorized: true,
                tx_code_constraints: Some(tx_code_constraints),
            } => Some(tx_code_constraints.clone()),
            Authorization {
                pre_authorized: true,
                tx_code_constraints: None,
            } => None,
            Authorization {
                pre_authorized: false, ..
            } => None,
        };
```

which you can then further simplify to:
```rust
        let tx_code_constraints = if authorization.pre_authorized {
            authorization.tx_code_constraints.clone()
        } else {
            None
        };
```

agent_api_rest/src/issuance/offers/mod.rs
```diff
@@ -45,6 +46,8 @@ pub(crate) async fn offers(
         // Unreachable error
         .ok_or_else(|| ApiError::new(StatusCode::INTERNAL_SERVER_ERROR))?;
 
+    let server_config = query_handler(SERVER_CONFIG_ID, &state.query.server_config).await?;
```
nanderstabel: You should not have to query `server_config` twice so this an indication that this function should be refactored a bit. 

if you first obtain the `credential_configurations:
```rust
    let credential_configurations = query_handler(SERVER_CONFIG_ID, &state.query.server_config)
        .await?
        .map(|server_config_view| server_config_view.credential_configurations) // Unreachable error
        .ok_or_else(|| ApiError::new(StatusCode::INTERNAL_SERVER_ERROR))?;
```

then you can use it to get the `persisted_credential_configuration_ids`:
```rust
    let persisted_credential_configuration_ids = credential_configurations.keys().collect::<Vec<_>>();
```

which can be used in the next block:
```rust
    if let Some(credential_configuration_id) = credential_configuration_ids.iter().find(|credential_configuration_id| {
        !persisted_credential_configuration_ids.contains(credential_configuration_id)
    }) {
        return Err(ApiError::builder(StatusCode::NOT_FOUND)
            .title("No Credential Configuration Found")
            .type_url(type_url("issuance#no-credential-configuration-found"))
            .message(format!(
                "No Credential Configuration found with id: `{credential_configuration_id}`"
            ))
            .finish());
    }
```

You can also make use of the `credential_configurations` to obtain the `tx_code_constraints`:
```rust
    let tx_code_constraints = credential_configuration_ids
        .iter()
        .find_map(|credential_configuration_id| {
            credential_configurations
                .get(credential_configuration_id)
                // TODO: find a way to bind the `authentication` to the Offer somehow
                .and_then(|(_, _, authorization)| {
                    if authorization.pre_authorized {
                        authorization.tx_code_constraints.clone()
                    } else {
                        None
                    }
                })
        });
```

Note that I changed something else as well: 
```rust
    let tx_code_constraints = credential_configuration_ids
        .first()
        .and_then(|credential_configuration_id| {
```

to 
```rust
    let tx_code_constraints = credential_configuration_ids
        .iter()
        .find_map(|credential_configuration_id| {
```

This is because if the Offer contains more than one Credential, then we should check if at least one of them requires a `tx_code` instead of just checking the first one.

It's not great honestly and this makes me think that `authentication` should not be part of the `CredentialConfiguration` struct, but rather bound to the Offer somehow. But, for now this current solution should be ok, we'll come up with a better design (hence why I added the `TODO` comment in there..

nanderstabel (changes requested):

agent_api_rest/src/issuance/credentials.rs
coplat: so clean!

agent_api_rest/src/issuance/offers/mod.rs
```diff
@@ -34,19 +35,16 @@ pub(crate) async fn offers(
...
+    let credential_configurations = query_handler(SERVER_CONFIG_ID, &state.query.server_config)
         .await?
-        .map(|server_config_view| {
-            server_config_view
-                .credential_configurations
-                .into_keys()
-                .collect::<Vec<_>>()
-        })
-        // Unreachable error
+        .map(|server_config_view| server_config_view.credential_configurations) // Unreachable error
```
nanderstabel: Can you please move the `// Unreachable error` comment to a newline?

agent_api_rest/src/issuance/offers/mod.rs
```diff
@@ -57,11 +55,28 @@ pub(crate) async fn offers(
             .finish());
     }
 
+    let tx_code_constraints = credential_configuration_ids
+        .iter()
+        .find_map(|credential_configuration_id| {
+            credential_configurations
+                .get(credential_configuration_id)
+                // TODO: find a way to bind the `authentication` to the Offer somehow
```
nanderstabel: `authentication` should be `authorization` 🙈

nanderstabel (changes requested):

nanderstabel (approved):

PR #216, opened 2025-09-02
fix: initialize database clients

daniel-mader:
# Description of change

Initialize the database clients (MongoDB, Postgres) just once and reuse them to reduce the amount of open connections.

n/a

n/a

### DIFF
--- a/agent_application/src/main.rs
+++ b/agent_application/src/main.rs
@@ -41,24 +41,48 @@ async fn main() -> io::Result<()> {
         vec![Box::new(EventPublisherHttp::load().unwrap())];
 
     let (identity_state, issuance_state, holder_state, verification_state) = match config().event_store.type_ {
-        EventStoreType::Postgres => (
-            agent_store::identity_state::<Postgres>(identity_services, identity_event_publishers).await,
-            postgres::issuance_state(issuance_services, issuance_event_publishers).await,
-            agent_store::holder_state::<Postgres>(holder_services, holder_event_publishers).await,
-            agent_store::verification_state::<Postgres>(verification_services, verification_event_publishers).await,
-        ),
-        EventStoreType::MongoDb => (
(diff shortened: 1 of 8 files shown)


agent_application/src/main.rs
```diff
@@ -42,22 +42,40 @@ async fn main() -> io::Result<()> {
 
     let (identity_state, issuance_state, holder_state, verification_state) = match config().event_store.type_ {
         EventStoreType::Postgres => (
-            agent_store::identity_state::<Postgres>(identity_services, identity_event_publishers).await,
+            agent_store::identity_state::<Postgres>(&Postgres, identity_services, identity_event_publishers).await,
```
nanderstabel: nit: Now all the `_state` functions expect `builder: &CCB` the type checker doesn't need the `::<Postgres/MongoDb/InMemory>` 'hints' anymore so you could omit those to make the formatting of this match block a bit leaner.

agent_store/src/mongodb.rs
```diff
@@ -28,10 +28,24 @@ where
...
+
+impl MongoDB {
+    pub async fn new() -> Self {
+        let connection_string = config().event_store.connection_string.clone().expect(
+            "Missing config parameter `event_store.connection_string` or `UNICORE__EVENT_STORE__CONNECTION_STRING`",
+        );
+        let client = default_mongo_client(&connection_string).await;
+        Self { client }
+    }
+}
```
nanderstabel: https://docs.rs/sqlx-core/0.8.6/sqlx_core/pool/struct.Pool.html

> `Pool` is Send, Sync and Clone. It is intended to be created once at the start of your application/daemon/web server/etc. and then shared with all tasks throughout the process’ lifetime.

Looks like `Postgres` would benefit from this approach as well. Do you think it would make sense to update the `Postgres` builder as well in this PR?

nanderstabel (approved):

Issue #202, opened 2025-07-24
Template management


daniel-mader:
### Description

In order to issue credentials more efficiently and to provide more guidance for users, a credential templating system is proposed.

### Data model
The data model for a template should include:
- **id**: unique identifier, generated by UniCore
- **name**: display name for easy reference
- **description**: (optional) description of what the template is used for and what it contains
- ~~**version**: number meant to be incremented by UniCore on every change of the **data** (not metadata)~~
- **logo_url**: (optional) a URL to the logo of the credential
- **creator**
  - **name**: display name of the creator of the template
  - **url**: URL of the credential issuer who created the template
- **holder_type**: either _individual_ or _organization_ to indicate whether this template is meant for B2C or B2B issuance
- **created_at**: should never change after initial creation
- **last_modified_at**: a timestamp when the template was edited or a new version was pulled from the original creator
- **tags**: (optional) a set of strings to allow quick reference for searching and grouping, limitations (such as max. length of each tag or the total number of allowed tags are to be discussed)
- **data**
  - **template**: a json object compliant with a credential data model (such as W3C v1.1, Open Badges 3.0, etc.) which also contains **placeholders** in the form of strings following the mustache convention (example: "first_name": "{{FIRST_NAME}}")
  - **placeholders**: JSON Schema where the "properties" are used to fill the **placeholders** in the **template**. No business restrictions need to be enforced in UniCore, only a check if the JSON Schema is according to standard.

### Commands
- **AddTemplate** (providing the full data model as described above, except the values generated by UniCore)
- **GetTemplates** (optional: filter by **holder_type**)
- **UpdateTemplate** (providing any part of the data model above will patch the existing data)
- **RemoveTemplate**
- **CloneTemplate** (copies the entire object, generates a new UUID, appends "Copy" to the **name**)

### HTTP API endpoints
- `/v0/templates`
  - **POST**: **AddTemplate**, **DuplicateTemplate** (indicated via a body such as `{"duplicate_from": "<uuid>"}`)
  - **GET**: **GetTemplates**
  - **DELETE**: **RemoveTemplate**
  - **PATCH**: **UpdateTemplate**

### Further requirements & considerations

1) (Cheyenne) Add raw issuance in dev mode:
Toggle raw issuance in dev mode: Flag `SKIP_TEMPLATE_ENFORCEMENT` which defaults to false.
2) (Cheyenne) Add RemoveTemplate and DuplicateTemplate endpoints
[...]

PR #221, opened 2025-09-30
feat: introduce credential template manager

nanderstabel:
# Description of change
This pull request introduces the new `agent_library` domain, which will be responsible for managing shareable credential `Templates`.

**Note:** I discovered that the HTTP Event Publisher was missing some events (which means it could not publish certain events), so this PR also fixes that. 

- https://github.com/impierce/ssi-agent/issues/202

n/a


agent_event_publisher_http/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition.workspace = true
 rust-version.workspace = true
 
 [dependencies]
+agent_library = { path = "../agent_library" }
```
daniel-mader: Should this be alphabetically sorted as well?

agent_identity/Cargo.toml
```diff
@@ -56,14 +56,14 @@ agent_shared = { path = "../agent_shared", features = ["test_utils"] }
 agent_store = { path = "../agent_store" }
 
 axum.workspace = true
+async-std = { version = "1.5", features = ["attributes", "tokio1"] }
```
daniel-mader: Shouldn't this be even before axum?

agent_store/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition.workspace = true
 rust-version.workspace = true
 
 [dependencies]
+agent_library = { path = "../agent_library" }
```
daniel-mader: This should be lower in the list (if alphabetical sorting is respected here).

daniel-mader (changes requested):
When I try to run this code as usual (`cargo run --release`), it tries to read from `../agent_shared/tests/test.config.yaml`. I suppose somewhere the feature `test_utils` is turned on which tries to load the test config. In the current `beta`, everything works as intended.

agent_identity/Cargo.toml
nanderstabel: True.. --> fixed (and replaced it with `async-std.workspace = true`)

agent_event_publisher_http/Cargo.toml
nanderstabel: fixed

agent_store/Cargo.toml
nanderstabel: fixed

nanderstabel:
> When I try to run this code as usual (`cargo run --release`), it tries to read from `../agent_shared/tests/test.config.yaml`. I suppose somewhere the feature `test_utils` is turned on which tries to load the test config. In the current `beta`, everything works as intended.

True, indeed I turned on the feature flag in one of the `Cargo.toml` files. It's fixed now :ok_hand:

agent_library/src/template/aggregate.rs
```diff
@@ -0,0 +1,540 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+pub struct Display {
+    pub name: String,
+    pub logo: Option<Logo>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+pub enum CredentialFormat {
+    #[serde(rename = "w3c_vc_data_model_v1-1")]
+    W3CVcDataModelV11,
```
daniel-mader, line 27: Should we add a link to the spec for each of the supported formats?

agent_api_rest/src/library/templates/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+};
+use http_api_problem::ApiError;
+use hyper::{header, StatusCode};
+use serde::{Deserialize, Serialize};
+use tracing::debug;
+
+#[derive(Deserialize, Serialize, Default)]
+#[serde(default, rename_all = "camelCase")]
+pub struct PostTemplatesEndpointRequest {
+    pub id: Option<String>,
```
daniel-mader: Are users allowed to provide their own `id`? I believe it can make sense for some use cases, such as restoring a template without breaking links. I'm asking, because usually we don't do this, but we generate the id for internal purposes only.

nanderstabel: I removed it just to be consistent. We can always make it an optional field again when there is an explicit reason for it.

daniel-mader (approved):

PR #224, opened 2025-10-03
feat: implement NATS event publisher 

coplat:
# Description of change
This PR implements a new NATS event publisher crate. 
The publisher converts domain events to CloudEvents and publishes them to NATS subjects for consumption by other services (e.g., our email delivery microservice). It can be configured to listen to particular events through the config file.  


agent_api_rest/src/issuance/offers/mod.rs
```diff
@@ -152,6 +155,9 @@ pub mod tests {
                         serde_json::to_vec(&json!({
                             "offerId": OFFER_ID,
                             "credentialConfigurationIds": ["001"],
+                            "deliveryOptions": {
+                                "recipient_email": null
+                            }
```
nanderstabel: We wouldn't want to force the client to explicitly state there is no `recipient_email` both because it is inconvenient but also because it would be a breaking change without a valid reason.

So to solve this you need to add the `#[serde(default)]` attribute to the `#[serde(default)]`  in the `OffersEndpointRequest` struct. You can then also remove:
```rust
                            "deliveryOptions": {
                                "recipient_email": null
                            }
```

from all the tests again.

agent_issuance/src/offer/aggregate.rs
```diff
@@ -104,6 +111,9 @@ impl Aggregate for Offer {
 
                 // If TxCode constraints are provided, generate a transaction code.
                 let tx_code = tx_code_constraints.as_ref().map(generate_tx_code);
+                if tx_code.is_some() && delivery_options.recipient_email.is_none() {
+                    return Err(MissingRecipientEmailError);
+                }
```
nanderstabel: The `tx_code.is_some() && delivery_options.recipient_email.is_none()` should not be a an invalid scenario. 

It would break the following (valid) flow:
1. UniCore Client wants UniCore to create an Offer that utilizes a Transaction Code (but without delivery option)
2. UniCore CLient sends GET request to the `/v0/offers/<offer_id>`  endpoint to obtain the Transaction Code`
3. UniCore CLient delivers the Transaction Code themselves to the Wallet Holder
4. Wallet Holder receives the Transaction Code and enters it when asked for it in their Wallet
5. The Issuance flow continues as usual.

In essence, the (email) delivery service is an optional service, not required.

agent_event_publisher_nats/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+            .id(event_id)
+            .source(event_source)
+            .ty(format!("offer.event.{}", event_type.to_lowercase()))
+            .data("application/json", serde_json::to_value(event)?)
+            .build()?;
+
+        // Convert Cloudevent into a suitable NATS message format
+        let nats_event = NatsCloudEvent::from_event(cloud_event)?;
+
+        println!(
```
nanderstabel: this should be `info!`

nanderstabel (changes requested):

agent_issuance/src/offer/aggregate.rs
coplat: good point, didn't think about this. will remove that block

agent_application/example.config.yaml
```diff
@@ -36,6 +36,12 @@ event_publishers:
     events:
       server_config: []
       credential: [UnsignedCredentialCreated, CredentialSigned]
+  nats:
+    enabled: true
+    nats_url: "nats://localhost:4222"
+    subject: "email.commands"
```
daniel-mader: I suggest we already abstract to a list of subject objects, like this:

```
nats:
  subjects:
    - name: "email.commands"
      events:
        offer: [TxCodeGenerated]
    - name: "system.users.data.verifications.example"
        credential: [CredentialSigned]    
```

agent_event_publisher_nats/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+edition.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+agent_issuance = { path = "../agent_issuance" }
+agent_shared = { path = "../agent_shared" }
+agent_store = { path = "../agent_store" }
+
+anyhow.workspace = true
+async-nats = "0.42"
```
daniel-mader, line 13: A newer version is available, so you can possibly bump it.

agent_shared/src/config/mod.rs
```diff
@@ -567,6 +568,14 @@ pub struct EventPublisherHttp {
     pub events: Events,
 }
 
+#[derive(Debug, Deserialize, Clone, Default, Serialize)]
+pub struct EventPublisherNats {
+    pub enabled: bool,
+    pub nats_url: String,
+    pub subject: String,
```
daniel-mader: See comment in **agent_application/example.config.yaml** regarding the support of multiple subjects.

agent_application/example.config.yaml
coplat: added handling for this ^

agent_event_publisher_nats/Cargo.toml
coplat: This causes dependency conflicts with nuid (and rand crates). Let's wait for an upstream fix?

coplat: Resolved :)

agent_api_rest/src/issuance/offers/send.rs
```diff
@@ -8,26 +8,29 @@ use axum::{
...
+    }): Json<SendOfferEndpointRequest>,
 ) -> Result<Response, ApiError> {
     let command = OfferCommand::SendCredentialOffer {
         offer_id: offer_id.clone(),
-        target_url,
+        delivery_method,
     };
 
-    // Send the Credential Offer to the `target_url`.
+    // Send the Credential Offer to the `target_url` or to the email recipient.
```
nanderstabel: nit: `... to the recipient's email`.

agent_issuance/src/offer/aggregate.rs
```diff
@@ -91,6 +116,14 @@ impl Aggregate for Offer {
...
                 // If TxCode constraints are provided, generate a transaction code.
                 let tx_code = tx_code_constraints.as_ref().map(generate_tx_code);
+                if tx_code.is_some()
+                    && delivery_options
+                        .as_ref()
+                        .and_then(|opts| opts.recipient_email.as_ref())
+                        .is_none()
+                {
+                    return Err(MissingRecipientEmailError);
+                }
```
nanderstabel: Do we really want to enforce that the tx code can only be sent via email? Why would we not want to allow it that clients would create an offer that includes a TX code first, then they query (read) the Offer aggregate through the api and then deliver the tx code via their own means?

agent_issuance/src/offer/aggregate.rs
```diff
@@ -172,28 +217,54 @@ impl Aggregate for Offer {
...
+                            status: Status::Pending,
+                        }])
+                    }
+                    DeliveryMethod::Email { recipient_email } => {
+                        info!("Sending credential offer via email to: {}", recipient_email);
+
+                        let offer_link = config()
+                            .application_url
+                            .join(&format!("offer/{}", offer_id))
+                            .expect("Failed to construct offer link URL");
```
nanderstabel, line 249: What is the purpose of the `offer_link`? Here the `offer/{}` path segment is appended to the `application_url` but we do not have an `/offer` endpoint 🤔 We do have an `/v0/offers` endpoint though, is that the one that's supposed to be used here? And who will call this `offer_link`?

agent_issuance/src/offer/aggregate.rs
```diff
@@ -343,6 +421,7 @@ pub mod tests {
                 credential_configuration_ids: vec![],
                 grant_types: grant_types.clone(),
                 tx_code_constraints: None,
+                delivery_options: None,
```
nanderstabel, line 416: I think it would be good to have at least one test to include a `Some()` value. You already added a `delivery_options` fixture below but it is unused atm.

agent_issuance/src/offer/aggregate.rs
```diff
@@ -40,6 +41,29 @@ pub struct Offer {
     pub credential_response: Option<CredentialResponse>,
     pub status: Status,
     pub tx_code: Option<String>,
+    #[serde(default)]
```
nanderstabel: This is not really required right or is it? 🤔

agent_issuance/src/offer/command.rs
```diff
@@ -7,16 +8,17 @@ use oid4vci::{
...
-#[serde(untagged)]
+#[serde(tag = "type")]
 pub enum OfferCommand {
     CreateCredentialOffer {
         offer_id: String,
         credential_configuration_ids: Vec<String>,
         grant_types: Vec<GrantType>,
         tx_code_constraints: Option<TxCodeConstraints>,
+        #[serde(default)]
+        delivery_options: Option<DeliveryOptions>,
```
nanderstabel: What is the reason you had to add/update these `serde` attributes? Just curious because would that mean that we also need to change `#[serde(untagged)]` to `#[serde(tag = "type")]` for the other Commands?

On another note, I actually think none of the Commands actually need to be de-serializable at all because because unlike events they are not persisted and I see no other reason why it is necessary to deserialize from a json object 🤔

nanderstabel (changes requested):

agent_issuance/src/offer/aggregate.rs
nanderstabel: If I am not mistaken this is the same issue as [this one](https://github.com/impierce/ssi-agent/pull/224#discussion_r2413776238) right?

agent_issuance/src/offer/aggregate.rs
coplat: The `offer_link` here was generated solely to point to the page which we now host as our so-called "launchpad" for receiving offers through an email, since most email clients were removing the non-https offer urls.

agent_issuance/src/offer/aggregate.rs
coplat: Yeah you're right. I mistakenly had in mind that if not via email, there would be no other way to send off the tx_code...

agent_issuance/src/offer/aggregate.rs
nanderstabel: Ok but the `application_url` is an 'internal' URL (in contrast to the `public_url`), and there is no `/offer` endpoint/route defined in `agent_api_rest` so how does that work? :thinking:

nanderstabel (approved):

daniel-mader (approved):

PR #236, opened 2025-11-10
feat: use IOTA Sponsoring Service to fund `did:iota` transactions

nanderstabel:
# Description of change

* Adds `iota_sponsoring_service` and `iota_sponsoring_service_auth` config variable that can be used to (pre) configure the client for the IOTA Sponsoring Service.
* Adds `/.well-known/sponsoring-configuration` endpoint which exposes data (most importantly iota wallet address) that IOTA Sponsoring Service can use to register UniCore as a client.

**Edit:**
* changed `iota_sponsoring_service` ==> `iota_sponsoring_service_url`
* changed `/.well-known/sponsoring-configuration` ==> `/public/sponsoring-configuration`

n/a

Can be tested against the Moveathon IOTA Sponsoring Service demo deployment.


agent_api_http/src/v0/identity/mod.rs
```diff
@@ -37,5 +40,6 @@ pub fn router(identity_state: IdentityState) -> Router {
         )
         .route("/.well-known/did.json", get(did))
         .route("/.well-known/did-configuration.json", get(did_configuration))
+        .route("/.well-known/sponsoring-configuration", get(sponsoring_configuration))
```
daniel-mader: Do you think it's a good idea to expose this as a `.well-known` even if it's not standardized in any way? We don't have to put it under `.well-known`, just because we want to access it from the outside. I also think this could even be challenging from a security perspective, since all `.well-known` endpoints are exposed to the public by default and you might not want to do that by default for your organisational crypto address?
We could put it under `/public/sponsoring-configuration` or something.

daniel-mader (approved):

PR #242, opened 2025-12-01
feat: duplicate and delete templates

coplat:
# Description of change
Initialize commands and endpoints for `DuplicateTemplate` and `DeleteTemplate`, allowing duplication of existing templates which results in a template with a new ID and the word "Copy" appended to the end of the template's title. 

To duplicate a template you can now make a `POST` request to `v0/templates/duplicate-template` with the body:
```
{
    "sourceTemplateId": "example-template-id-123"
}
```

To delete a template you can simply make a `POST` request to `v0/template/delete-template with the body: 
```
{
    "templateId": "example-template-id-123"
}
```
Additionally, the previous `PATCH` request has been now replaced by a `POST` request to `v0/templates/update-template` to update elements of a template.

Our postman collection and OpenAPI documentation have been updated accordingly. 

# BREAKING CHANGE
- List all `GET /templates` is now under `GET /templates/get-all-templates`
- Create `POST /templates` is now under  `POST /templates/create-template`  
- Update `PATCH /templates/{template_id}` has moved to `POST /templates/update-template`. 

Addresses parts of Issue #202 

All existing tests pass. Functionality has also been tested through the postman collection. 


agent_library/src/template/command.rs
```diff
@@ -62,4 +63,11 @@ pub enum TemplateCommand {
         template_id: String,
         schema: serde_json::Value,
     },
+    DeleteTemplate {
+        template_id: String,
+    },
+    DuplicateTemplate {
+        duplicate_from: String,
+        new_template_id: String,
+    },
```
nanderstabel: This one we don't really need anymore right since you now added `duplicate_from` to `CreateTemplate`?

agent_library/src/template/event.rs
```diff
@@ -76,6 +77,13 @@ pub enum TemplateEvent {
         schema: serde_json::Value,
         modified_at: String,
     },
+    TemplateDeleted {
+        template_id: String,
+    },
+    TemplateDuplicated {
+        duplicate_from: String,
+        new_template_id: String,
```
nanderstabel: Similar to https://github.com/impierce/ssi-agent/pull/242/files#r2589651162, we don't need this one right?

agent_library/src/template/views/mod.rs
```diff
@@ -128,6 +130,10 @@ impl View<Template> for Template {
                 self.schema.replace(schema.clone());
                 self.modified_at.replace(modified_at.clone());
             }
+            TemplateDeleted { template_id: _ } => {
+                self.status = Status::Deleted;
```
nanderstabel, line 134: Similar to the code you added to the `apply` function for this Event in `aggregate.rs`:
```rust
                *self = Self::default();
                self.template_id = template_id;
                self.status = Status::Deleted;
```

Here all the values should also be 'wiped'.

agent_api_rest/src/library/templates/mod.rs
```diff
@@ -12,10 +12,13 @@ use http_api_problem::ApiError;
 use hyper::{header, StatusCode};
 use serde::{Deserialize, Serialize};
 use tracing::debug;
+use uuid::Uuid;
 
 #[derive(Deserialize, Serialize, Default)]
 #[serde(default, rename_all = "camelCase")]
 pub struct PostTemplatesEndpointRequest {
+    pub id: Option<String>,
```
nanderstabel: Since you branched off from my previous PR before it got merged I removed the `id` field from this request struct so you can also remove it again.

agent_api_rest/src/library/templates/mod.rs
```diff
@@ -46,10 +51,53 @@ pub(crate) async fn post_templates(
...
 ) -> Result<Response, ApiError> {
-    let template_id = uuid::Uuid::new_v4().to_string();
+    // Handle template duplication if `duplicate_from` field is provided.
+    if let Some(old_template_id) = duplicate_from {
+        let new_template_id = Uuid::new_v4().to_string();
+
+        // Fetch original template data.
+        let original_template = query_handler(&old_template_id, &state.query.template)
+            .await?
+            .ok_or_else(|| ApiError::new(StatusCode::NOT_FOUND))?;
```
nanderstabel: Returning a 404 NOT FOUND status code would communicate to the client that the endpoint itself does not exist, which is not true. I think it is better to use [422 Unprocessable Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/422) here and include a Problem Details. You can follow this example in `credentials.rs` (which also uses `NOT_FOUND` which is wrong so please add a TODO comment there 😅 ):
```rust
            ApiError::builder(StatusCode::NOT_FOUND)
                .title("No Credential Configuration Found")
                .type_url(type_url("issuance#no-credential-configuration-found"))
                .message(format!(
                    "No Credential Configuration found with id: `{credential_configuration_id}`"
                ))
                .finish()
```

What this means as well is that you will need to add the description of the error as well to the `docs` folder. You will need to add a `library.md` file and include the description (you can take a look at the Problem Details descriptions we already have in the other `.md` files for inspiration).

agent_api_rest/src/library/templates/mod.rs
```diff
@@ -46,10 +51,53 @@ pub(crate) async fn post_templates(
...
+            display: original_template.display.clone(),
+            credential_format: original_template.credential_format.clone(),
+            creator: original_template.creator.clone(),
+            holder_type: original_template.holder_type.clone(),
+            tags: original_template.tags.clone(),
+            status: Status::Draft,
+            visibility: original_template.visibility.clone(),
+            description: original_template.description.clone(),
+            r#type: original_template.r#type.clone(),
+            schema: original_template.schema.clone(),
```
nanderstabel: If I am not mistaken you can remove all the `.clone()` occurrences here.

nanderstabel (changes requested):

agent_api_http/postman/ssi-agent.postman_collection.json
```diff
@@ -1981,6 +1982,23 @@
...
 							"name": "Create a new Template",
+							"event": [
+								{
+									"listen": "test",
+									"script": {
+										"exec": [
+											"if (pm.response.code === 201) {",
+											"    const response = pm.response.json();",
+											"    pm.collectionVariables.set(\"template_id\", response.id);",
+											"    console.log(\"Saved template_id: \" + response.id);",
```
nanderstabel: Here as well as in the Request Body for "Duplicate Template" you use `template_id` but for the other endpoints we use `TEMPLATE_ID`. Is that deliberate or can we use `TEMPLATE_ID` for everything?

agent_api_http/postman/ssi-agent.postman_collection.json
```diff
@@ -2380,6 +2448,10 @@
 			"key": "REQUEST_URI",
 			"value": "INITIAL_VALUE",
 			"type": "string"
+		},
+		{
+			"key": "template_id",
+			"value": ""
```
nanderstabel: Can we re-use `TEMPLATE_ID` instead of adding this one?

agent_api_rest/src/library/templates/mod.rs
```diff
@@ -46,10 +51,53 @@ pub(crate) async fn post_templates(
...
+        let new_template_id = Uuid::new_v4().to_string();
+
+        // Fetch original template data.
+        let original_template = query_handler(&old_template_id, &state.query.template)
+            .await?
+            .ok_or_else(|| ApiError::new(StatusCode::NOT_FOUND))?;
+
+        let create_command = TemplateCommand::CreateTemplate {
+            template_id: new_template_id.clone(),
+            duplicate_from: Some(old_template_id),
```
daniel-mader: Although I called it **duplicate_from** in issue https://github.com/impierce/ssi-agent/issues/202, I think that name it isn't fitting too well (apart from that it's grammatically wrong as well -- should be "duplicate of").

When you duplicate a template, then adapt it and change it quite a bit over time, something such as **original_template** is probably more fitting. This implies a duplicate, but more elegantly.

agent_library/src/template/aggregate.rs
```diff
@@ -62,6 +63,7 @@ pub enum Visibility {
 pub struct Template {
     #[serde(rename = "id")]
     pub template_id: String,
+    pub duplicate_from: Option<String>,
```
daniel-mader: See comment https://github.com/impierce/ssi-agent/pull/242/files#r2591724162

agent_library/src/template/command.rs
daniel-mader: I think we could also do it the other way: not introduce **duplicate_from** to **CreateTemplate** and have a separate **DuplicateTemplate** command, which implicitly just copies all fields. Otherwise, you'd have to supply all other fields as well when calling **CreateTemplate** (same **display**, same **credential_format**, etc.).

agent_library/src/template/event.rs
daniel-mader: See reply https://github.com/impierce/ssi-agent/pull/242/files#r2591738067

agent_library/src/template/command.rs
nanderstabel: I think that what you suggest would indeed be the right DDD way of doing this but atm technically that would be still a bit tricky to do. When the `Template::handle` function receives the command:
```rust
    DuplicateTemplate {
        duplicate_from: String,
        new_template_id: String,
    }
```

then it has no way of accessing the field data from `duplicate_from`. What we would need then is define a `LibraryService` that acts as a Domain Service. This Domain Service then should contain access to the `LibraryState` and it should have a method that simply accepts a Template ID, and returns the `TemplateView` which can be used in the `Template::handle` function to do the duplication. However, the problem here is that if we want to do this this will require some refactor related to `agent_store` (which I have started btw but not completed yet).

But once we have that then I think it's a good to idea to implement the `LibraryService` so what we can do what you suggested. So, `TODO` comment for now? :sweat_smile:

coplat: I agree that this would be a bit tricky for the moment. I added a `TODO` for now.

agent_library/src/template/event.rs
```diff
@@ -8,7 +8,7 @@ use strum::Display;
 pub enum TemplateEvent {
     TemplateCreated {
         template_id: String,
-        // TODO: Make this a required field.
```
daniel-mader, line 11: Was this TODO removed by accident? `title` still is an `Option<>` and I believe the TODO was meant for this field.

agent_api_http/src/v0/library/mod.rs
```diff
@@ -16,8 +21,14 @@ pub fn router(library_state: Arc<LibraryState>) -> Router {
         .nest(
             API_VERSION,
             Router::new()
-                .route("/templates", get(get_templates).post(post_templates))
-                .route("/templates/{template_id}", get(get_template).patch(patch_template)),
+                // GET requests
```
daniel-mader: I don't think we need those comments here. They don't really add anything to explain the code underneath and their behavior is already in the name of the `get()` function itself.

agent_api_http/src/v0/templates/mod.rs
```diff
@@ -49,14 +51,14 @@ impl From<Template> for TemplateDto {
...
-            schema: value.schema,
+            schema: *value.schema,
         }
     }
 }
 
 #[derive(Deserialize, Serialize, Default)]
 #[serde(default, rename_all = "camelCase")]
-pub struct PostTemplatesEndpointRequest {
+pub struct PostCreateTemplateEndpointRequest {
```
daniel-mader: We can drop the `Post` prefix here as well, so just `CreateTemplate...` reads better.

agent_api_http/src/v0/templates/mod.rs
```diff
@@ -121,9 +124,64 @@ pub(crate) async fn post_templates(
         .ok_or_else(|| ApiError::new(StatusCode::INTERNAL_SERVER_ERROR))
 }
 
+#[derive(Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PostDuplicateTemplateEndpointRequest {
```
daniel-mader: See https://github.com/impierce/ssi-agent/pull/242/files#r2689948478

agent_library/src/template/event.rs
coplat: You're right 😅

agent_issuance/src/credential/aggregate.rs
```diff
@@ -115,213 +115,234 @@ impl Aggregate for Credential {
...
+                let notification_id = test_utils::notification_id();
+                #[cfg(not(feature = "test_utils"))]
+                let notification_id = agent_shared::generate_random_string();
 
-                    let credential_subject = identity_credential::credential::Subject::from_json_value(
-                        data.raw["credentialSubject"].clone(),
-                    )
-                    .map_err(|e| InvalidCredentialSubjectError(e.to_string()))?;
+                // Check if template enforcement is to be skipped.
+                let skip_template_enforcement = std::env::var("SKIP_TEMPLATE_ENFORCEMENT").is_ok();
```
daniel-mader: We should be reading this from the `config()`, not directly from the `std::env`.

daniel-mader: Also, it can be inlined with the following `if` statement.

agent_library/src/template/views/mod.rs
daniel-mader: If we never return a template anywhere that has `Status::Deleted`, why would we "wipe" the values? That would prevent a template from being restored, right? We could introduce a "trash" in a future iteration - unless we "wipe" the data.

agent_api_http/src/v0/templates/mod.rs
```diff
@@ -265,9 +324,10 @@ pub(crate) async fn get_templates(
             let filtered_templates: Vec<TemplateDto> = all_templates_view
                 .templates
                 .into_values()
-                .filter(|_template|
+                .filter(|template| {
+                    template.status != Status::Deleted
```
daniel-mader, line 330: Instead of checking for the status here, I think deleted templates should not even be included in the `all_templates_view` any longer, so this check would be obsolete.

agent_api_http/src/v0/templates/mod.rs
```diff
@@ -285,41 +345,33 @@ pub(crate) async fn get_template(
 ) -> Result<Response, ApiError> {
     query_handler(&template_id, &state.query.template)
         .await?
-        .map(|template_view| (StatusCode::OK, Json(TemplateDto::from(template_view))).into_response())
+        .and_then(|template_view| {
+            if template_view.status == Status::Deleted {
```
daniel-mader, line 351: I'm not sure if we need this check here in the API layer since the `template_view` should not exist any longer if the template has been deleted.

agent_api_http/src/v0/templates/mod.rs
daniel-mader: We can also keep it as is for now, but we have to make sure to never return it anywhere.

coplat: I'll add a TODO

daniel-mader: @nanderstabel Do you think we should work on this now? If we leave it as is and we merge it with the TODO, we leave it up to the user to interpret the `template.status` correctly.

agent_shared/src/config/mod.rs
daniel-mader, line 1: I think you can undo those format changes.

agent_library/src/template/views/mod.rs
nanderstabel: yeah you're right that if we want to enable the option to restore it easily that we shouldn't wipe the data (so then actually we also shouldn't wipe it in the `apply` function for this Event in `aggregate.rs`).

agent_api_http/src/v0/templates/mod.rs
nanderstabel: With this filter in place the User will never get to see the deleted templates right so in that aspect it won't matter if we merge it as is or not, or am I misunderstanding?

Apart from that yes we can remove the view from the `all_template_view` (which would indeed mean we woudn't have to filter here). For 'individual' `template_view`s ([referring to your other comment below](https://github.com/impierce/ssi-agent/pull/242#discussion_r2693269187)) we cannot do that because the underlying `ViewRepository` trait only allows for loading and updating Views so if the Client/User still has the ID of the deleted View it would technically be possible to query the view (if we wouldn't use the filter). 

Difference with `all_<type>s_view` and `<type>_view`:
```
|         | `all_types_view`        | `type_view` | 
|---------|-------------------------|-------------|
| id      | "all_types_view"        | UUID        |
| content | HashMap<UUID, TypeView> | TypeView    |
```

where in both cases the actual record cannot be deleted (but in the latter case the specific `TypeView` can be removed from the HashMap). So if we would want to work on this now depends a bit on what you mean with " leave it up to the user to interpret " because how I understand it currently the User will not be able to see the deleted Templates :thinking:

agent_api_http/src/v0/templates/mod.rs
nanderstabel: Through `cqrs-es`'s `ViewRepository` trait Views can not be deleted actually, so the only way to determine whether a View is "deleted" is again through the `status` field.

docs/problem-details/library.md
```diff
@@ -0,0 +1,3 @@
+## Source Template Not Found
+
+This error arises when the source template data for duplication can not be fetched. This could indicate that the `template_id` used for duplication is not associated with an existing template. It can also arise in instances where the template has been deleted.
```
nanderstabel: The corresponding endpoint expects `sourceTemplateId` (camelCase) so I'd suggest to use that instead of `template_id`.

agent_api_http/src/v0/templates/mod.rs
daniel-mader: Okay, since we can't work around that shortcoming of the library right now, we can keep filtering out based on the status. @coplat Could you remove the `TODO` then since it's obsolete if we keep the filter?

daniel-mader (approved):

nanderstabel (approved):

PR #244, opened 2025-12-03
feat: initialize `IssuerMetadataSynchronizationPolicy`

nanderstabel:
# Description of change
This PR introduces the `IssuerMetadataSynchronizationPolicy` in the Issuance domain that will listen for specific `TemplateEvent`s from the Library Domain. In a near future update this policy will dispatch `ServerConfigCommand`s in order to update the Credential Issuer Metadata in order to keep the `credential_configurations_supported` field in sync with the Templates stored in `agent_library`.

In order to facilitate this, the Policy needs a reference to the `IssuanceState`:
```rust
pub struct IssuerMetadataSynchronizationPolicy {
    // TODO: Actually use this.
    _issuance_state: Arc<IssuanceState>,
}
```

This meant that `ApplicationState` cannot have sole ownership anymore over `IssuanceState` and therefore it has been wrapped into an `Arc`. The result of that is that many files in `agent_api_rest` had to be updated in order to wrap the state into `Arc`s.

This PR also includes a fix for the HTTP Response of the `GET /v0/templates/` and  `GET /v0/templates/<template_id>` endpoints that ensures that the responses are returned in camelCase.

- #202

* Since this merely an initialization of the Policy the only real 'proof' of the code working is that it compiles. The actual workings of the Policy will be implemented in https://github.com/impierce/ssi-agent/pull/245
* camelCase responses are tested through `test_template_response_serialization`.

### DIFF
--- a/agent_api_rest/src/identity/mod.rs
+++ b/agent_api_rest/src/identity/mod.rs
@@ -6,6 +6,8 @@ pub mod well_known;
 
 pub mod error;
 
+use std::sync::Arc;
+
 use agent_identity::state::IdentityState;
 use axum::{
     routing::{get, post},
@@ -21,7 +23,7 @@ use crate::{
     API_VERSION,
 };
 
-pub fn router(identity_state: IdentityState) -> Router {
+pub fn router(identity_state: Arc<IdentityState>) -> Router {
     Router::new()
         .nest(
             API_VERSION,
--- a/agent_api_rest/src/identity/mod.rs
+++ b/agent_api_rest/src/identity/mod.rs
@@ -1,3 +1,4 @@
+// Endpoint handlers
 pub mod connections;
 pub mod documents;
 pub mod profiles;
@@ -5,8 +6,6 @@ pub mod services;
 pub mod well_known;
 
 pub mod error;
-
-use std::sync::Arc;
 
 use agent_identity::state::IdentityState;
 use axum::{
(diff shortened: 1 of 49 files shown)


agent_api_rest/src/identity/mod.rs
```diff
@@ -6,6 +6,8 @@ pub mod well_known;
 
 pub mod error;
```
coplat, line 9: should there be a blank line between pub mod error and the other modules? and also between std::sync::Arc and the rest?

agent_api_rest/src/library/templates/mod.rs
```diff
@@ -11,8 +11,49 @@ use axum::{
...
+use std::sync::Arc;
 use tracing::debug;
 
+/// Data transfer object for Templates.
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TemplateDto {
+    #[serde(rename = "id")]
+    pub template_id: String,
+    pub title: Option<String>,
```
coplat, line 23: The field`duplicate_from` has been added to the Template struct in the aggregate so should probably be added here too.

nanderstabel: `duplicate_from` would be introduced in [this](https://github.com/impierce/ssi-agent/pull/242) PR which is not merged yet. Once your PR is merged and I rebase this branch on the new beta then it makes more sense to update any code here that is required then.

agent_api_rest/src/identity/mod.rs
nanderstabel: The empty line between `pub mod error` and the other modules above it is intentional because all those other modules specifically all contain endpoint handlers. I'd say that's not particularly useful especially because I did not add any comment to point that out, so I have now :+1: 

The other empty line under `use std::sync::Arc` was not intentional so I removed it :)

daniel-mader (approved):

PR #252, opened 2025-12-15
feat: initialize API layer for key management system

coplat:
# Description of change
Please write a summary of your changes and why you made them.


agent_api_http/src/v1/identity/keys/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+        .await
+        .map_err(|e| {
+            tracing::error!("Saga failed: {:?}", e);
+            ApiError::new(StatusCode::INTERNAL_SERVER_ERROR)
+        })?;
+
+    Ok((StatusCode::CREATED, managed_key_id))
+}
+#[derive(Deserialize)]
+pub struct PostRemoveKey {
```
nanderstabel, line 61: This struct also needs: `rename_all = "camelCase"` (same goes for `PostRenameAlias`)

agent_api_http/src/v1/identity/keys/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+    let view = query_handler("all_managed_keys", &context.state.query.all_managed_keys)
+        .await
+        .map_err(|e| {
+            tracing::error!("Query failed: {:?}", e);
+            ApiError::new(StatusCode::INTERNAL_SERVER_ERROR)
+        })?;
+
+    // AllManagedKeysView contains a HashMap of ManagedKeyViews
+    let keys = view
+        .unwrap()
```
nanderstabel: `view` will be `None` when there is no event recorded yet for `all_managed_keys` and in that case the endpoint should return `200 OK` with an empty vector as the payload.

agent_api_http/src/v1/identity/keys/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+#[derive(Deserialize)]
+pub struct PostRemoveKey {
+    pub key_id: String,
+}
+
+pub(crate) async fn remove_key(
+    State(context): State<IdentityContext>,
+    Json(payload): Json<PostRemoveKey>,
+) -> Result<StatusCode, ApiError> {
+    let managed_key_id = query_handler(&payload.key_id, &context.state.query.managed_key)
```
nanderstabel: We cannot use `payload.key_id` for the `view_id` unfortunately because `key_id` is not the Aggregate identifier (`managed_key_id` is). 

So what we need is a method to map this the `key_id` to the corresponding `managed_key_id` and the only way of doing that currently is by calling `query_handler("all_managed_keys", &context.state.query.all_managed_keys)` and then `find` the correct key in the `AllManagedKeysView`. It would be nice if we had a generic method for this that would also work elsewhere but for now I would suggest you create a function that you can also reuse in the other endpoints that have the same problem (`rename_key_alias`, `set_signing_key`).

agent_api_http/src/v1/identity/keys/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+    let keys = view
+        .unwrap()
+        .managed_keys
+        .into_values() // Iterate over the 'ManagedKeyView' items inside
+        .map(|key_view| ManagedKeyDto {
+            managed_key_id: key_view.managed_key_id,
+            key_id: key_view.key_id,
+            alias: key_view.alias,
+            signing_algorithm: key_view.signing_algorithm,
+        })
```
nanderstabel: This endpoint should filter out all the views which `is_removed` field is set to `false`, e.g:
```rust
     .filter_map(|key_view| {
            (!key_view.is_removed).then(|| ManagedKeyDto {
                managed_key_id: key_view.managed_key_id,
                key_id: key_view.key_id,
                alias: key_view.alias,
                signing_algorithm: key_view.signing_algorithm,
            })
        })
```

nanderstabel (approved):

PR #261, opened 2025-12-31
feat: configure IOTA node URL and Basic Auth

Oran-Dan:
# Description of change
Since did-manager repo has also been updated to allow for configuring the IOTA node url used by the Resolver, this repo also needed a few small updates to allow for the updates in did-manager to be used in this repo.

Blocked by:
- https://github.com/impierce/did-manager/pull/41

After the blocking PR is merged also the Cargo.toml refs need to be updated in this PR

A new unit test has been added.
Old tests still pass.

### DIFF
--- a/agent_secret_manager/src/subject.rs
+++ b/agent_secret_manager/src/subject.rs
@@ -57,9 +57,9 @@ impl Subject {
         &mut self,
         tls_config: Option<rustls::ClientConfig>,
         node_url: Option<NodeUrls>,
-        set_user_password: Option<(&str, &str)>,
+        basic_auth: Option<(&str, &str)>,
     ) {
-        self.resolver = Resolver::new_with_options(tls_config, node_url, set_user_password).await;
+        self.resolver = Resolver::new_with_options(tls_config, node_url, basic_auth).await;
     }
 
     pub async fn get_public_key(&self, key_id: KeyId, algorithm: &Algorithm) -> anyhow::Result<Jwk> {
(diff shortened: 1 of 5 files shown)


agent_secret_manager/src/subject.rs
```diff
@@ -318,4 +343,26 @@ mod tests {
         let public_key = UnparsedPublicKey::new(&ED25519, public_key_bytes);
         assert!(public_key.verify(message.as_bytes(), &signature_bytes).is_ok());
     }
+
+    #[tokio::test]
+    async fn test_configure_resolver_node_urls() {
```
daniel-mader: This test is not asserting anything and even if it did, it should not try to connect to an external resource (such as the IOTA testnet), because that would mean the tests would fail if the network cannot be reached for whatever reason.

agent_secret_manager/src/subject.rs
```diff
@@ -21,19 +22,46 @@ use tokio::sync::Mutex;
...
+
         Self {
             stronghold_storage,
             verification_method_ids: Arc::new(Mutex::new(HashMap::new())),
+            resolver: Resolver::new_with_options(None, Some(node_urls), username_password).await,
         }
     }
 
+    /// If no node URLs are provided, the resolver will use the default IOTA node URLs.
+    pub async fn configure_resolver(
```
daniel-mader: This function is only used in a test and can be removed since the resolver is configured implicitly when calling `did_manager_consumer::resolver::Resolver::new_with_options()`.

daniel-mader (approved):

PR #269, opened 2026-01-12
feat: initialize Nonce endpoint 

coplat:
# Description of change
This PR initializes the nonce endpoint as described in OID4CI's [draft 15 specification. ](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-ID2.html#name-nonce-endpoint). This is vital as a means to prevent replay attacks during credential requests. Nonces are generated when a post request is made to the `/oid4vci/nonce` endpoint, returning a unique and fresh `c_nonce`.

This `c_nonce` must then be included in the proof of a CredentialRequest made to the issuer. If that nonce has already been redeemed or does not exist, the credential request is denied. 

If it is valid, the c_nonce is then marked as redeemed and can no longer be reused. 

This PR initializes the nonce endpoint which generates the nonce, and validates the nonce when a credential request is made within the API layer. Further, the nonce commands, events, errors and corresponding aggregate are also initialized within this PR. 

- Blocked by https://github.com/impierce/openid4vc/pull/115

Existing tests have been updated to include nonces. Nonce endpoint has been added to the postman collection for `c_nonce` generation. 


agent_issuance/src/state.rs
```diff
@@ -141,6 +148,7 @@ pub async fn load_server_metadata(state: &IssuanceState) -> anyhow::Result<()> {
                 credential_issuer_metadata: Box::new(CredentialIssuerMetadata {
                     credential_issuer: public_url.clone(),
                     credential_endpoint: public_url.append_path_segment("openid4vci/credential"),
+                    nonce_endpoint: Some(public_url.append_path_segment("openid4vci/nonce")),
```
daniel-mader, line 151: I think this should also be included in the OpenAPI, right?

agent_api_http/src/v0/issuance/nonce/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+    let fresh_c_nonce = generate_random_string();
+    let command = NonceCommand::GenerateNonce {
+        c_nonce: fresh_c_nonce.clone(),
+    };
+
+    command_handler(&fresh_c_nonce, &state.command.nonce, command)
+        .await
+        .map_err(|_| ApiError::new(StatusCode::INTERNAL_SERVER_ERROR))?;
+
+    Ok((StatusCode::OK, Json(json!({ "c_nonce": fresh_c_nonce }))).into_response())
```
daniel-mader: [According to the spec](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-ID2.html#name-nonce-response), we also need to require to include the HTTP header `Cache-Control: no-store` for security reasons. Please also consider writing a test for this.

daniel-mader (changes requested):

agent_issuance/src/state.rs
coplat: Yes it is

agent_api_http/src/v0/issuance/nonce/mod.rs
coplat: you're right, i missed that. added the headers + test

agent_api_http/postman/ssi-agent.postman_collection.json
```diff
@@ -846,6 +846,33 @@
...
+						"header": [
+							{
+								"key": "Authorization",
+								"value": "Bearer {{ACCESS_TOKEN}}"
+							}
+						],
+						"body": {
+							"mode": "raw",
+							"raw": "{\n    \"credential_configuration_id\": \"001\",\n    \"proof\": {\n        \"proof_type\": \"jwt\",\n        \"jwt\": \"eyJ0eXAiOiJvcGVuaWQ0dmNpLXByb29mK2p3dCIsImFsZyI6IkVkRFNBIiwia2lkIjoiZGlkOmtleTp6Nk1rdWlSS3ExZktyekFYZVNOaUd3cnBKUFB1Z1k4QXhKWUE1Y3BDdlpDWUJEN0IjejZNa3VpUktxMWZLcnpBWGVTTmlHd3JwSlBQdWdZOEF4SllBNWNwQ3ZaQ1lCRDdCIn0.eyJpc3MiOiJkaWQ6a2V5Ono2TWt1aVJLcTFmS3J6QVhlU05pR3dycEpQUHVnWThBeEpZQTVjcEN2WkNZQkQ3QiIsImF1ZCI6Imh0dHA6Ly8xOTIuMTY4LjEuMTI3OjMwMzMvIiwiZXhwIjo5OTk5OTk5OTk5LCJpYXQiOjE1NzEzMjQ4MDAsIm5vbmNlIjoidW5zYWZlX2Nfbm9uY2UifQ.wR2e4VUnVjG6IK9cntcqvc_8KEJQUd3SEjsPZwDYDlYqijZ4ZaQLxyHtzNmLkIS3FpChLrZrcvIUJrZxrWcKAg\"\n    }\n}"
+						},
```
nanderstabel: The spec [does not require a Authorization header and request body](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#name-nonce-request) so I believe this may be an unintended result from a copy-paste from another Request?

agent_api_http/src/v0/issuance/nonce/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+
+use axum::Json;
+use http_api_problem::ApiError;
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+use std::sync::Arc;
+
+#[derive(Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct NonceEndpointRequest {}
```
nanderstabel: This is not used anywhere so it can be removed I presume?

agent_api_http/postman/ssi-agent.postman_collection.json
```diff
@@ -846,6 +846,33 @@
 						}
 					},
 					"response": []
+				},
+				{
+					"name": "nonce",
```
nanderstabel: With the introduction of the Nonce endpoint the nice issuance 'flow' in the Postman Collection breaks. Previously we could use scripts to for example obtain the `ACCESS_TOKEN`, `NOTIFICATION_ID`, etc.. which then could be used further on in the issuance flow. For the `C_NONCE` this would technically still be possible but too cumbersome at this point and not really a sensible use of the Postman Collection  (because we would need to implement a Proof signer which embeds the `nonce` in Javascript). 

Still, just to imply that there is some order of which the endpoints are called during the issuance flow I would suggest to move this part up above the `credential` request in line 637:

```json
                # here
				{
					"name": "credential",
					"event": [
                    ...
```

agent_api_http/src/v0/issuance/credential_issuer/credential.rs
```diff
@@ -40,6 +59,33 @@ pub(crate) async fn credential(
...
+                        .map_err(|_| PublicError::from(CredentialErrorResponse::InvalidProof))?;
+                }
+                None => {
+                    // Nonce doesn't exist
+                    return Err(PublicError::from(CredentialErrorResponse::InvalidNonce));
+                }
+            }
+        }
+    }
+
```
nanderstabel: Since this is more Application/Domain Logic, would you say it would make sense to wrap this logic (including `extract_nonce_from_proof`) into a new Application Service e.g. `NonceValidationService`, similar to the `AccessTokenValidationService` we already have in `agent_issuance/src/application/access_token_validation_service.rs`?

agent_api_http/src/v0/issuance/credential_issuer/credential.rs
```diff
@@ -266,9 +313,18 @@ pub mod tests {
 
     pub async fn credential(
         issuance_app: &mut Router,
+        issuance_state: &Arc<IssuanceState>,
         access_token: String,
         external_server: Option<MockServer>,
     ) -> (String, String) {
+        let test_nonce = "7e03ad3f76cb3338c3a5642fe7634476aa3ad93fa1d584011ba2150d9da47133";
```
nanderstabel: nit: consider adding a `TEST_NONCE` `const` to line 224 that you can use here and in line 418.

nanderstabel (changes requested):

agent_api_http/src/v0/issuance/credential_issuer/credential.rs
coplat: good point!

nanderstabel (dismissed):

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+  "[json]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode",
+    "editor.formatOnSave": true
```
daniel-mader: I agree with suggesting a default formatter for certain files types (`json`), but running `formatOnSave` for every user of this repository can be considered "intrusive".
You can always keep these settings for your personal work environment without checking them into the code, so I suggest removing this file from this PR. 👍

agent_api_http/postman/ssi-agent.postman_collection.json
daniel-mader, line 1: Formatting changes unrelated to the scope of this PR. Please undo and open a separate "refactoring" PR which contains purely formatting changes, if the team agrees.

agent_issuance/src/nonce/aggregate.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    async fn handle(
+        &self,
+        command: Self::Command,
+        _services: &Self::Services,
+    ) -> Result<Vec<Self::Event>, Self::Error> {
+        use NonceCommand::*;
+        use NonceEvent::*;
+        // use NonceError::* Unused for now - aggregate operations don't fail
```
daniel-mader: Is this a TODO?

agent_event_publisher_http/src/lib.rs
```diff
@@ -49,6 +50,7 @@ pub struct EventPublisherHttp {
     pub server_config: Option<AggregateEventPublisherHttp<ServerConfig>>,
     pub credential: Option<AggregateEventPublisherHttp<Credential>>,
     pub offer: Option<AggregateEventPublisherHttp<Offer>>,
+    pub nonce: Option<AggregateEventPublisherHttp<Nonce>>,
```
daniel-mader, line 53: Just a random question: in what scenario do we need to publish a nonce to a subscriber? Should the nonce validation logic ever leave UniCore? What can subscribers do with the nonce if they receive it?

agent_issuance/src/application/nonce_validation_service.rs
```diff
@@ -0,0 +1,75 @@
...
+            claims.get("nonce").and_then(|n| n.as_str()).map(|s| s.to_string())
+        }
+    }
+}
+
+impl NonceValidationService {
+    /// Validates a nonce included in the credential request.
+    ///
+    /// This function checks if the nonce exists and whether it has already been redeemed.
+    pub async fn validate(
```
daniel-mader: Could you write some unit tests for this function?

agent_issuance/src/nonce/aggregate.rs
```diff
@@ -0,0 +1,69 @@
...
+use tracing::{debug, info};
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct Nonce {
+    pub c_nonce: String,
+    pub is_redeemed: bool,
+}
+
+#[async_trait]
+impl Aggregate for Nonce {
```
daniel-mader, line 18: Could you write some tests for this using `cqrs_es::test::TestFramework` just to verify that events are correctly processed and updating the state?

agent_store/src/lib.rs
```diff
@@ -526,7 +537,6 @@ pub(crate) fn partition_event_publishers(event_publishers: Vec<Box<dyn EventPubl
             if let Some(access_token) = event_publisher.access_token() {
                 partitions.access_token_event_publishers.push(access_token);
             }
-
```
daniel-mader, line 529: Is there a reason why the empty lines are removed here, but not further up or was that unintentional?

agent_issuance/src/nonce/aggregate.rs
daniel-mader: This is especially crucial if we introduce new event versions so we have assertions that different event versions are processed as intended.

agent_event_publisher_http/src/lib.rs
coplat: Code won't compile without an aggregate being in the EventPublisherHttp, so leaving this in there for now.

agent_store/src/lib.rs
coplat: This whole block was inconsistent in regards to line spacing so I've removed the 3 inconsistent lines.

daniel-mader: Could you then be consistent in removing the inconsistencies and also remove the blank line in L561 or is there a reason why this one should stay?

agent_api_http/postman/ssi-agent.postman_collection.json
Oran-Dan: I would agree but it seems that just because of this we are now adding the .prettierignore file above, only to be removed in the next refactor? In that case I would be against it tbh. That feels like introducing tech-debt even though it makes this PR less easy to review. It seems that our current approach to tech-debt will create this situation nearly every PR. We're going to fix tech-debt "along the way", this results in PR's containing changes somewhat unrelated to the scope.

agent_issuance/src/application/nonce_validation_service.rs
```diff
@@ -0,0 +1,75 @@
...
+        state: &IssuanceState,
+        credential_request: &CredentialRequest,
+    ) -> Result<(), NonceValidationError> {
+        if let Some(proof) = &credential_request.proof {
```
Oran-Dan: I think this line should be moved to the helper function `extract_nonce_from_proof` which would then have to be renamed to `extract_nonce_from_credential_request`. I think so because the `proof` variable is used nowhere in this function but only passed to the helper so cleaner to only initialize this variable where it's actually used.

Furthermore, please keep helper functions at the bottom of the file.

agent_api_http/postman/ssi-agent.postman_collection.json
Oran-Dan: Feel free to resolve again if you don't agree with my comment. I unresolved mainly because otherwise this comment doesn't even show up.

daniel-mader: Postman changed their formatting (indentation level) in one of their recent updates. Since it is a generated file, we don't want to constantly have to apply our own formatters on top of it and we just want to take it "as is" and never actually touch it in any way. That's why it is a conscious decision to exclude it from Prettier. We also chose to use a folder-specific prettierignore because we did not see any benefit in putting this on the project's root level since it's a "one-off".

agent_store/src/lib.rs
coplat: I think there's a weird formatting thing going on here. In my VSCode the spacing is consistent, ie there is no space in line 561

daniel-mader: Maybe it shifted since there were more changes in the file.

daniel-mader (approved):

PR #273, opened 2026-01-19
test: replace blocking `Subject::default()` with async initialization

nanderstabel:
# Description of change
Since the addition of `pub resolver: Resolver,` to `Subject` in [this commit](https://github.com/impierce/ssi-agent/commit/d42a8f9a7336bb7eec8f45e0eb969c12e34e7f0e#diff-49ea6b0853e5c54834add274451694740b725f16d35cdecd6b5d3fdfee4e3377) we're initializing `Resolver` once and then reuse it (which is a good thing). Somehow this causes many tests to timeout due to a `block_on` in `impl Default for Subject` that would cause an upstream `jsonrpsee` crate to get 'stuck'.

This PR fixes that by replacing the Default impl for a custom async `test_subject` method. 

This change caused one of the tests to overflow:
```bash
thread 'offer::aggregate::tests::test_verify_credential_response' (718784) has overflowed its stack
fatal runtime error: stack overflow, aborting
error: test failed, to rerun pass `-p agent_issuance --lib`

Caused by:
  process didn't exit successfully: `/home/nander/Impierce/ssi-agent/target/debug/deps/agent_issuance-3f00251efa88b62a test_verify_credential_response --nocapture` (signal: 6, SIGABRT: process abort signal)
```

To fix that I changed `cargo test --workspace` to `RUST_MIN_STACK=16777216 cargo test --workspace` in `format-lint-test.yaml`

n/a

All tests run successfully now. 

### DIFF
--- a/.github/workflows/format-lint-test.yaml
+++ b/.github/workflows/format-lint-test.yaml
@@ -26,4 +26,6 @@ jobs:
         run: cargo clippy --all-targets --all-features -- -D warnings
 
       - name: Test
+        # TODO: We set `RUST_MIN_STACK` to 16 MiB to prevent stack overflow in `test_verify_credential_response`.
+        # Investigate if this is still needed after our testing refactor.
         run: RUST_MIN_STACK=16777216 cargo test --workspace
--- a/.github/workflows/format-lint-test.yaml
+++ b/.github/workflows/format-lint-test.yaml
@@ -26,4 +26,4 @@ jobs:
         run: cargo clippy --all-targets --all-features -- -D warnings
 
       - name: Test
-        run: cargo test --workspace
+        run: RUST_MIN_STACK=16777216 cargo test --workspace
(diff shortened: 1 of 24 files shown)


.github/workflows/format-lint-test.yaml
```diff
@@ -26,4 +26,4 @@ jobs:
         run: cargo clippy --all-targets --all-features -- -D warnings
 
       - name: Test
-        run: cargo test --workspace
+        run: RUST_MIN_STACK=16777216 cargo test --workspace
```
daniel-mader, line 31: Could you please leave a comment why it needs to be that exact number and how you got to it? If I'm not mistaken, this is 16 MiB. Maybe consider to make it a TODO since I believe we should refactor the test code so free up the stack so this wouldn't be necessary any more.

daniel-mader (approved):

PR #290, opened 2026-02-18
docs: generate OpenAPI specification

daniel-mader:
# Description of change

This change initiates the generation of an OpenAPI specification directly from the Rust code itself.

- This is a start, so only a few endpoints are covered in this PR. It should evolve over time.
- For the transition period (until the entire API documentation is fully generated), we'll have two files:
  - The current `openapi.yaml` (hand-written) stays untouched and still used as the primary reference.
  - The new `openapi-generated.yaml` (generated) should grow over time until it surpasses the hand-written version.

n/a

A new test generates the `openapi-generated.yaml` file on each run. A GitHub Action checks if there are diffs compared to the current version. This effectively fails the pipeline if a developer forgot to re-generate the file.


agent_api_http/src/v0/templates/mod.rs
```diff
@@ -16,8 +16,10 @@ use std::sync::Arc;
 use tracing::debug;
 use uuid::Uuid;
 
+pub mod openapi;
+
 /// Data transfer object for Templates.
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, utoipa::ToSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct TemplateDto {
```
coplat, line 24: I think we can use here something like `#[schema(as = Template)]` to prevent it from being called TemplateDto in the final doc

agent_api_http/src/v0/identity/connections/openapi.rs
```diff
@@ -0,0 +1,11 @@
+use crate::v0::identity::connections::{__path_get_connection, __path_get_connections};
+use utoipa::OpenApi;
+
+#[derive(OpenApi)]
+#[openapi(
+    paths(get_connections, get_connection),
+    tags(
+        (name = "connections", description = "Manage trusted connections."),
```
coplat: Could we make the C in 'connections' uppercase? I think it would look better in the final doc.

agent_api_http/src/v0/templates/openapi.rs
```diff
@@ -0,0 +1,14 @@
+use crate::v0::templates::{
+    __path_create_template, __path_duplicate_template, __path_get_template, __path_get_templates,
+};
+use utoipa::OpenApi;
+
+#[derive(OpenApi)]
+#[openapi(
+    paths(create_template, duplicate_template, get_templates, get_template),
+    tags(
+        (name = "library", description = "Manage your own templates, browse and import external templates."),
```
coplat: Same thing here as in this [comment](https://github.com/impierce/ssi-agent/pull/290#discussion_r2826533975) - I think having the tags uppercased is more aligned with how we currently have our openapi.yaml formatted.

agent_api_http/src/v0/templates/mod.rs
daniel-mader: Awesome find! I used `title =` first, which didn't work. `as =` seems to work! 👍

agent_api_http/src/v0/identity/connections/openapi.rs
daniel-mader: I just followed the [official reference](https://swagger.io/docs/specification/v3_0/grouping-operations-with-tags/) where they use lowercase, but I'm not too opinionated about it. I agree with the fact that it looks nicer when loaded into a tool such as Scalar. 👍

nanderstabel (approved):

PR #291, opened 2026-02-23
feat: enable ELM issuance, add Json Schema validation

Oran-Dan:
# Description of change
This PR enables the issuance of ELM credentials both in JwtVcJson format and in vc+sd-jwt format. 

It also refactors the issuance flow in `CreateUnsignedCredential` and `SignCredential` to be clearer and more scalable as we add ELM to the list of Credential Data Models which we can issue.

Furthermore, this PR adds the Json Schema validation folder, which includes the relevant Json Schemas and the structs and functions to make use of them.

NOTE: I need to add unit tests for ELM in `agent_issuance/src/credential/aggregate.rs` but its definitely reviewable already since I did very extensive live testing.

- 

Old tests still pass.
New tests have been added for the `json_schema` folder.
I've not yet been able to add unit tests for the sd-jwt issuance cases.
This is because the salts keep changing of course but our testcode actually lives in the `cqrs` crate. 
So to refactor the code to allow sd-jwt testing means opening another PR in that crate as well, which at this point seems to much of a delay for this PR. 
Curious to hear the reviewers opinions on this? (keep in mind live tests have covered the same code as described below)

The following credential issuance flows have been live tested against UniMe on the `fix/vcdm2` branch (which is ready to be merged):
- jwt_vc_json VC 1.1
- jwt_vc_json ELM
- jwt_vc_json OBv3
- dc+sd-jwt
- vc+sd-jwt VC 2.0
- vc+sd-jwt ELM
- vc+sd-jwt OBv3


agent_library/src/json_schema/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+pub trait ValueToString {
+    fn to_clean_string(&self) -> Option<String>;
+}
+
+impl ValueToString for serde_json::Value {
+    /// A simple helper function to convert a `serde_json::Value` to an `Option<String>`.
+    /// The original as_str or to_string methods work terribly due to including quotes characters.
+    /// The original as_str/to_string methods output the following: "/".../"" or Some("/".../"").
+    /// This function cleanly outputs Some("...").
+    /// Renaming this clarifies our code instead of having as_str and to_string calls everywhere.
```
nanderstabel: I believe the description below is a bit more accurate:
```rust
    /// Helper to convert a `serde_json::Value` to an owned `Option<String>`.
    /// 
    /// This resolves two common inconveniences:
    /// 1. `to_string()` serializes the JSON, resulting in extra quotes (e.g., "\"value\"").
    /// 2. `as_str()` returns a `&str`, but often an owned `String` is required.
    /// 
    /// Returns `Some(String)` if the value is a JSON string, or `None` otherwise.
```

agent_library/src/json_schema/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+}
+
+// Structs
+
+#[derive(serde::Deserialize)]
+#[serde(untagged)]
+pub enum StringOrArray {
+    String(CredentialType),
+    Array(Vec<CredentialType>),
+}
```
nanderstabel: This code is not being used anywhere. Do you think we can leave it out for now and only (re)introduce it once we do need to use it?

agent_library/src/json_schema/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+        .build(json_schema)
+        .map_err(|_| {
+            JsonSchemaError::InvalidJsonData(format!(
+                "Failed to compile JSON Schema from serde_json::Value: {json_schema}"
+            ))
+        })
+}
+
+/// This struct is solely used to implement the `Retrieve` trait from the `jsonschema` crate,
+/// allowing us to reference other JSON Schema's in this folder via $ref in any JSON Schema.
```
nanderstabel: nit: Since `MemoryRetriever` doesn't contain any fields you can use a unit struct syntax here: `struct MemoryRetriever {}` --> `struct MemoryRetriever;`



(also nit: `JSON Schema's` --> `JSON Schemas`)

agent_library/src/json_schema_validation.rs
```diff
@@ -0,0 +1,424 @@
...
+    #[error("Get Credential Type Error: `{0}`")]
+    GetCredentialTypeError(String),
+    #[error("Failed to parse JSON data: `{0}`")]
+    InvalidJsonData(String),
+    #[error("Credential data validation, according to its type `{0}`, results in the following errors: `{1:#?}`")]
+    CredentialValidationError(String, Vec<String>),
+}
+
+#[cfg(test)]
+mod tests {
```
nanderstabel, line 238: All the tests below test `validate_credential_types` but that function does not seem to be used anywhere. So, should there maybe tests for the code that we do use elsewhere instead?

agent_issuance/src/credential/aggregate.rs
```diff
@@ -813,6 +1043,62 @@ fn get_status_list_url(index: usize) -> Result<identity_core::common::Url, Crede
     Ok(status_list_url.into())
 }
 
+// Helper methods to simplify working with serde_json::Value.
+pub trait ExtraMethods {
+    /// Inserts a value at the specified path, creating intermediate objects as needed.
+    /// The path includes the final key name where the value will be inserted.
+    /// For example, to set `$.issuer.id = "123"`, use:
+    /// `credential.add_value_or_insert(&["issuer", "id"], json!("123"))`
```
nanderstabel: `add_value_or_insert` does not seem to correspond with the trait methods

agent_issuance/src/credential/aggregate.rs
```diff
@@ -800,6 +584,452 @@ impl Aggregate for Credential {
...
+                            "https://www.w3.org/ns/credentials/v2",
+                            "https://purl.imsglobal.org/spec/ob/v3p0/context-3.0.3.json"
+                        ]),
+                    )
+                    .ok_or(BuildCredentialError(
+                        "Failed to enter the @context into the credential".to_string(),
+                    ))?;
+
+                // If no root id is provided, convert the aggregate credential_id to an urn as sensible default.
+                println!("1credential_id: {credential_id:?}");
```
nanderstabel: Can be removed I assume?

agent_issuance/src/credential/aggregate.rs
```diff
@@ -813,6 +1043,62 @@ fn get_status_list_url(index: usize) -> Result<identity_core::common::Url, Crede
...
+
+        // Insert the value at the final key if it doesn't exist
+        current_value
+            .as_object_mut()?
+            .entry(last_key.to_string())
+            .or_insert(value);
+
+        Some(self)
+    }
+}
```
nanderstabel: Do you think these may be generic enough to move them to `agent_shared`?

nanderstabel: Also, perhaps a more appropriate name for the trait would be `ValueExt` or `SerdeJsonValueExt`. You could then also add `to_clean_string` in there so we just have one specific trait with 'tools' to work with serde_json::Value.

nanderstabel (changes requested):

agent_library/src/json_schema/mod.rs
Oran-Dan: you're right, sorry already discussed that but completely forgot to change it

agent_library/src/json_schema/mod.rs
Oran-Dan: Yeah verification side is only going to use this fn, I can reintroduce it when implementing it there.

agent_library/src/json_schema/mod.rs
Oran-Dan: renamed it as suggested by Daniel and added a test for it too

agent_library/src/json_schema_validation.rs
Oran-Dan: that fn does use all the other code inside mode.rs so it was full coverage. Changed to testing the validate() method directly now

agent_issuance/src/credential/aggregate.rs
Oran-Dan: thanks!

agent_issuance/src/credential/aggregate.rs
Oran-Dan: yes

agent_issuance/src/credential/aggregate.rs
Oran-Dan: yes I agree, self issuance in UniMe could profit from this as well

agent_library/src/json_schema_validation.rs
nanderstabel: ah ok you're right I missed that :ok_hand:

nanderstabel (approved):

PR #330, opened 2026-04-28
feat(templates): enforce template usage, validate against `schema`

daniel-mader:
# Description of change
- Templates are required for credential issuance.
- Credential configurations are manged implicitly through the templates.
- Nested structures (such as in Open Badges 3.0) are flattened.
- Some properties in the JSON Schema can be marked as "immutable" through `schema_properties_attributes`, so they are locked and can never be altered/removed.

### BREAKING CHANGE
Require `templateId` when creating a credential. The two fields `data_model` and `holder_type` become immutable after the initial template creation.


agent_api_http/src/v0/library/error.rs
```diff
@@ -15,6 +15,31 @@ impl IntoApiErrorExt for TemplateError {
                 .type_url(type_url("templates#invalid-schema-properties-attributes"))
                 .source(self)
                 .finish(),
+            TemplateError::NonRemovablePropertyViolation(_) => ApiError::builder(StatusCode::CONFLICT)
```
coplat, line 23: These errors should be added to the problem details documentation. The type_url's should be "library#..." instead of "templates#".

agent_shared/src/config/mod.rs
```diff
@@ -701,9 +701,7 @@ pub enum TemplateEvent {
     TemplateCreated,
     TitleUpdated,
     DisplayUpdated,
-    DataModelUpdated,
     CreatorUpdated,
-    HolderTypeUpdated,
```
nanderstabel, line 726: The event types can now also be removed from `ssi-agent/agent_event_publisher_http/README.md`

agent_issuance/src/application/credential_configuration_projection.rs
```diff
@@ -0,0 +1,505 @@
...
+    /// Re-queries the current template state from the view repository and, if the template is not
+    /// in Draft status, updates (or creates) the corresponding credential configuration.
+    async fn sync_from_view(&self, template_id: &str) {
+        use agent_library::template::aggregate::Status;
+
+        let Some(view) = self.template_view.get() else {
+            warn!("Template view not yet initialized; skipping credential configuration sync for `{template_id}`");
+            return;
+        };
+        match query_handler(template_id, view).await {
```
coplat, line 57: The match arm skips Draft and Deleted but implicitly allows Published and Archived. Should this be intended for Archived?

agent_library/src/template/aggregate.rs
```diff
@@ -527,6 +598,282 @@ fn validate_schema_properties_attributes(
...
+}
+
+/// Merges OpenBadges 3.0 default required properties into a user-provided schema.
+/// Ensures that the standard-mandated fields are always present and required.
+///
+/// # Panics
+/// Panics if `schema` is not a JSON object. Callers must ensure the schema is an object
+/// (or default to `{"type": "object"}`) before calling this function.
+#[cfg(test)]
+fn merge_open_badges_defaults(schema: &mut serde_json::Value) {
```
coplat: this test fn seems to only be called by `test_merge_open_badges_defaults_into_empty_schema`. but since the actual production code seems to be using `validate_open_badges_required_properties` as opposed to this auto-merge logic, I think this could be removed as it doesn't reflect the actual production code

agent_issuance/src/application/credential_configuration_projection.rs
```diff
@@ -0,0 +1,505 @@
...
+        _ => "vc+sd-jwt",
+    }
+    .to_string();
+
+    let type_ = match template.data_model {
+        DataModel::W3CVcDataModelV1_1 | DataModel::W3CVcDataModelV2_0 => {
+            vec!["VerifiableCredential".to_string()]
+        }
+        DataModel::OpenBadges3_0 => {
+            vec!["OpenBadgeCredential".to_string(), "AchievementCredential".to_string()]
```
nanderstabel: For OpenBadges, the `type` value should be an array where the first item must be `"VerifiableCredential"` and the second item must me either `"OpenBadgeCredential"` OR `"AchievementCredential"` (I think in most cases we default to  `"OpenBadgeCredential"`) (see: https://www.imsglobal.org/spec/ob/v3p0#achievementcredential)

agent_api_http/src/v0/library/error.rs
```diff
@@ -15,6 +15,31 @@ impl IntoApiErrorExt for TemplateError {
...
+                .source(self)
+                .finish(),
+            TemplateError::InvalidRequiredPropertyType(_) => ApiError::builder(StatusCode::BAD_REQUEST)
+                .title("Invalid Required Property Type")
+                .type_url(type_url("templates#invalid-required-property-type"))
+                .source(self)
+                .finish(),
+            TemplateError::MissingTitle => ApiError::builder(StatusCode::BAD_REQUEST)
+                .title("Missing Title")
+                .type_url(type_url("templates#missing-title"))
```
nanderstabel: All these `type_url`s are supposed to point to a specific Problem Details, but there is no description for them in the `/docs/problem-details/library.md` file

agent_api_http/src/v0/issuance/mod.rs
```diff
@@ -1,5 +1,4 @@
 // Endpoint handlers
-pub mod credential_configurations;
```
nanderstabel, line 2: Now this module is removed it is not included in the module tree anywhere, so that should mean you can now remove the `agent_api_http/src/v0/issuance/credential_configurations.rs` file.

agent_api_http/src/v0/issuance/credentials.rs
```diff
@@ -50,11 +55,11 @@ pub(crate) async fn credential(
 #[derive(Deserialize, Serialize, utoipa::ToSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct CredentialsEndpointRequest {
+    pub template_id: String,
     pub offer_id: String,
     pub credential: Value,
     #[serde(default)]
     pub is_signed: bool,
-    pub credential_configuration_id: String,
```
nanderstabel, line 57: Not only the Credentials Endpoint includes the `credential_configuration_id` but also the Offers Endpoint includes the `credential_configuration_ids` property (in `agent_api_http/src/v0/issuance/offers/mod.rs`), which is only used for the JIT Credentials flow. I think for consistency it should be removed there as well (or for now add a `TODO` comment so we can address that at a later point).

nanderstabel (changes requested):
Edit: nvm the comment below, I figured out I had to update the status to published :ok_hand: 

When I Create a new Template:
```curl
curl --location 'http://localhost:3033/v0/templates/create-template' \
--header 'Content-Type: application/json' \
--header 'X-API-KEY: {{API_KEY}}' \
--data '{
    "title": "My Example Template",
    "credentialFormat": "w3c_vc_data_model_v1-1",
    "holderType": "organization"
}'
```
I assumed that a corresponding Credential Configuration would be created and included to the Credential Issuer Metadata, but when I hit:
```curl
curl --location 'http://localhost:3033/.well-known/openid-credential-issuer' \
--header 'X-API-KEY: {{API_KEY}}'
```

No new Credential Configuration (with title `"My Example Template"`) is included in the Metadata. Is this expected, should I somehow trigger a Syncing command explicitly?

nanderstabel:
Since this PR makes makes Credential Configurations unusable in the API, it also does not make sense anymore to initialize those 'default' Credential Configurations on startup, so imo `fn update_credential_configurations` in `agent_issuance/src/state.rs` should be completely removed :partying_face:

agent_api_http/src/v0/issuance/credentials.rs
```diff
@@ -74,35 +79,96 @@ pub struct CredentialsEndpointRequest {
...
+        .map_err(|_| {
+            ApiError::builder(StatusCode::INTERNAL_SERVER_ERROR)
+                .title("Template Query Error")
+                .type_url(type_url("issuance#template-query-error"))
                 .message(format!(
-                    "No Credential Configuration found with id: `{credential_configuration_id}`"
+                    "An error occurred while looking up the template with id: `{template_id}`"
                 ))
                 .finish()
+        })?
```
nanderstabel: No need for this specific Problem Detail. You can just replace it with `?` and it will default to the generic implementation of `IntoApiErrorExt` for `PersistenceError`

agent_issuance/src/application/credential_configuration_projection.rs
```diff
@@ -0,0 +1,505 @@
...
+                    }
+
+                    let template = Template {
+                        template_id: template_id.clone(),
+                        title: title.clone(),
+                        display: *display.clone(),
+                        data_model: data_model.clone(),
+                        r#type: r#type.clone(),
+                        status: status.clone(),
+                        ..Default::default()
```
nanderstabel, line 310: There should be a line with `schema: schema.clone()` in here,[ otherwise build_claims_from_schema will always exit early on this line:
```rust
let schema = template.schema.as_ref().as_ref()?;
```

agent_issuance/src/application/credential_configuration_projection.rs
```diff
@@ -0,0 +1,507 @@
...
+
+    let claims: Vec<ClaimDescription> = properties
+        .keys()
+        .filter_map(|key| {
+            let path_elements: Vec<ClaimPathElement> = key
+                .split('.')
+                .map(|segment| ClaimPathElement::String(segment.to_string()))
+                .collect();
+
+            let path = ClaimPathPointer::try_new(path_elements).ok()?;
```
nanderstabel: If the provided properties look like this:
```json
"properties": {
            "age": {},
            "name": {}
}
```

then the claim path pointers will look like this:
- `["age"]`
- `[name"]`

but especially since this code is only executed when `format == "vc+sd-jwt"`, it's more likely that the claim pat pointers end up as:
- `["credentialSubject", "age"]`
- `["credentialSubject", name"]`

agent_issuance/src/application/credential_configuration_projection.rs
```diff
@@ -0,0 +1,507 @@
...
+                background_color: None,
+                text_color: None,
+            }])
+        });
+
+    let claims = if format == "vc+sd-jwt" {
+        build_claims_from_schema(template)
+    } else {
+        None
+    };
```
nanderstabel, line 162: Why only build claims when `format == "vc+sd-jwt"`? I think you can make it work for `"dc+sd-jwt"` as well if you rewrite this part to something like this:
```rust
    let claims = if format == "vc+sd-jwt" {
        build_claims_from_schema(template, "credentialSubject."])
    } else if format == "dc+sd-jwt" {
        build_claims_from_schema(template, "")
    } else {
        None
    };
```

where `"credentialSubject."` can be prepended to `key` to create the `path_elements` in `fn build_claims_from_schema`

related: https://github.com/impierce/ssi-agent/pull/330/changes#r3220225971

agent_issuance/src/application/credential_configuration_projection.rs
daniel-mader: We should indeed treat "Archived" as "read-only", but that has more implications. We're not actively supporting that at the moment, so we can leave it as is for now and properly cover it in an upcoming task. The main goal for now is to keep the credential configurations and template updates in sync.

agent_issuance/src/application/credential_configuration_projection.rs
```diff
@@ -0,0 +1,820 @@
...
+/// The display name is taken from `template.display.name` if present, falling back to `template.title`.
+/// When the format is "vc+sd-jwt", claims are derived from `schema.properties` merged with
+/// `schema_properties_attributes.selectivelyDisclosable`.
+///
+/// The `credential_definition.type` array prefers the template's explicit `type` values.
+/// Narrow fallbacks are only used when the template does not yet provide any type values.
+fn credential_configuration_from_template(template: &Template) -> CredentialConfiguration {
+    let format = match template.data_model {
+        DataModel::W3CVcDataModelV1_1 => "jwt_vc_json",
+        _ => "vc+sd-jwt",
```
coplat, line 106: In lines 156-159 below: the format `dc+sd-jwt` is mentioned. That format should probably also be listed here.

daniel-mader: This will be covered with full dc+sd-jwt support.

agent_api_http/src/v0/issuance/offers/mod.rs
```diff
@@ -23,50 +28,68 @@ use std::sync::Arc;
 #[serde(rename_all = "camelCase")]
 pub struct OffersEndpointRequest {
     pub offer_id: String,
-    #[serde(default)]
-    pub credential_configuration_ids: Vec<String>,
+    pub template_id: String,
```
nanderstabel: It was previously optional to provide `credentialConfigurationIds` because `credentialConfigurationId` (now `templateId`) is already required in the `/v0/credentials` endpoint, so does it need to be required here now? The reason `credentialConfigurationIds` was introduced here was to enable the JIT Credential flow (in that context this field IS indeed required (or rather conditionally required)).

Also `credentialConfigurationIds` was plural, allowing multiple JIT Credentials in one Offer. Can this be changed to `pub template_ids: Vec<String>,`?

agent_issuance/src/application/credential_configuration_projection.rs
```diff
@@ -0,0 +1,820 @@
...
+    } else {
+        None
+    };
+
+    CredentialConfiguration {
+        credential_configuration_id: template.template_id.clone(),
+        format,
+        type_,
+        credential_metadata: CredentialMetadata { display, claims },
+        authorization: Authorization::default(),
```
nanderstabel: `authorization: Authorization::default(),` basically means: always use Pre-Authorized Code flow, so you will need to add a property to the Template aggregate that can be used to set the Authorization to: 
```json
            "authorization": {
                "pre_authorized": false
            }
```

If `pre_authorized` is `false`, this triggers the Authorization Code flow (or the Interactive Flow if `UNICORE__ENABLE_INTERACTIVE_AUTHORIZATION_FLOW` is `true`)

agent_api_http/src/v0/issuance/credentials.rs
```diff
@@ -303,30 +625,288 @@ pub mod tests {
...
+            credential_configuration,
+            provisioned: false,
+        };
+
+        agent_shared::handlers::command_handler(SERVER_CONFIG_ID, &issuance_state.command.server_config, command)
             .await
             .unwrap();
 
+        template_id
+    }
```
nanderstabel: It looks as if this helper function was generated to circumvent a gap in the (templates) API just to satisfy existing tests. Especially the second part where it explicitly creates a Credential Configuration is the problem, because this behavior is impossible in a running application because the `/v0/credential-configurations` endpoint is removed. 

- The previous `/v0/credential-configurations` endpoint allowed to set `pre_authorized` to `false`
- This endpoint was removed, so through the API it is not possible to set `pre_authorized` to `false`
- This helper function circumvents that regression in the API by dispatching the `UpdateCredentialConfiguration` directly to the `command_handler`. 

(related comment: https://github.com/impierce/ssi-agent/pull/330/changes#r3468401582)

Instead of 'cheating' the test by dispatching the Commands directly to a command_handler, these kind of helper functions should always call the (HTTP) API (like how you did it in `fn unsigned_credentials_request_with_template`), because it their behavior should reflect how a Client would be able to use the API. If then, tests do not pass, it likely means there is a gap, or regression in the (management) API.

agent_api_http/src/v0/issuance/credentials.rs
```diff
@@ -303,30 +625,288 @@ pub mod tests {
...
+
+            let command = ServerConfigCommand::UpdateCredentialConfiguration {
+                credential_configuration,
+                provisioned: false,
+            };
+
+            agent_shared::handlers::command_handler(SERVER_CONFIG_ID, &issuance_state.command.server_config, command)
+                .await
+                .unwrap();
+        }
```
nanderstabel: See https://github.com/impierce/ssi-agent/pull/330/changes#r3468636124

agent_api_http/src/v0/authorization/authorization_server/token.rs
```diff
@@ -150,16 +156,13 @@ pub mod tests {
...
 
-        let mut app = issuance::router(issuance_state.clone());
+        let library_state = Arc::new(library_state(&InMemory, Default::default(), Default::default()).await);
+        create_test_template_with_auth(&library_state, &issuance_state, is_pre_authorized).await;
 
-        let credential_configuration_id = if is_pre_authorized {
-            "001".to_string()
-        } else {
-            "002".to_string()
-        };
```
nanderstabel, line 159: These Credential Configuration IDs mapped to those in the `test_credential_configurations.json` file. Since Templates now replaced Cred Configs in the API you can remove the following:
- the `test_credential_configurations.json` file
- Line 53 in `agent_shared/tests/test.config.yaml`
- Any mention of `credential_configuration_file` in `agent_shared/src/config/mod.rs`

agent_api_http/src/v0/issuance/offers/mod.rs
```diff
@@ -23,50 +28,68 @@ use std::sync::Arc;
...
+        template_id,
         delivery_options,
     }): Json<OffersEndpointRequest>,
 ) -> Result<Response, ApiError> {
-    // Check if the credential configuration IDs are valid.
-    let credential_configurations = query_handler(SERVER_CONFIG_ID, &state.query.server_config)
+    if template_id.is_empty() {
+        return Err(ApiError::builder(StatusCode::BAD_REQUEST)
+            .title("Missing Template ID")
+            .type_url(type_url("issuance#missing-template-id"))
```
nanderstabel: This error is missing a Problem Detail description in `docs/problem-details/issuance.md`. Also, since Credential Configurations are deprecated from the API, several existing Problem Detail descriptions need to be updated/removed in `docs/problem-details/issuance.md`

agent_api_http/src/v0/issuance/offers/mod.rs
```diff
@@ -23,50 +28,68 @@ use std::sync::Arc;
...
-            .title("No Credential Configuration Found")
-            .type_url(type_url("issuance#no-credential-configuration-found"))
-            .message(format!(
-                "No Credential Configuration found with id: `{credential_configuration_id}`"
-            ))
+        .filter(|t| t.status != TemplateStatus::Deleted)
+        .ok_or_else(|| {
+            ApiError::builder(StatusCode::UNPROCESSABLE_ENTITY)
+                .title("Template Not Found")
+                .type_url(type_url("issuance#template-not-found"))
```
nanderstabel: See https://github.com/impierce/ssi-agent/pull/330/changes#r3472472755

agent_api_http/src/v0/issuance/offers/mod.rs
```diff
@@ -23,50 +28,68 @@ use std::sync::Arc;
...
+                .message(format!("No template found with id: `{template_id}`"))
+                .finish()
+        })?;
+
+    if template.status != TemplateStatus::Published {
+        return Err(ApiError::builder(StatusCode::UNPROCESSABLE_ENTITY)
+            .title("Template Not Published")
+            .type_url(type_url("issuance#template-not-published"))
```
nanderstabel: See https://github.com/impierce/ssi-agent/pull/330/changes#r3472472755

agent_api_http/src/v0/issuance/offers/mod.rs
```diff
@@ -23,50 +28,68 @@ use std::sync::Arc;
...
+        .map(|(_, _, authorization)| authorization)
+        .ok_or_else(|| {
+            ApiError::builder(StatusCode::INTERNAL_SERVER_ERROR)
+                .title("No Credential Configuration Found")
+                .type_url(type_url("issuance#no-credential-configuration-found"))
+                .message(format!(
+                    "No Credential Configuration found with id: `{credential_configuration_id}`"
+                ))
+                .finish()
+        })?;
```
nanderstabel: This should probably become obsolete once `authorization` is stored in the Template aggregate

agent_api_http/src/v0/issuance/offers/mod.rs
```diff
@@ -79,11 +102,10 @@ pub(crate) async fn offers(
         GrantType::AuthorizationCode
     }];
 
-    // Create an offer if it does not exist yet.
     if query_handler(&offer_id, &state.query.offer).await?.is_none() {
         let command = OfferCommand::CreateCredentialOffer {
             offer_id: offer_id.clone(),
-            credential_configuration_ids,
+            credential_configuration_ids: vec![credential_configuration_id],
```
nanderstabel: See https://github.com/impierce/ssi-agent/pull/330/changes#r3468312827

agent_api_http/src/v0/issuance/credentials.rs
```diff
@@ -170,23 +243,139 @@ pub(crate) async fn credentials(
...
+) -> Result<(), ApiError> {
+    let actual_format = detect_signed_credential_format(signed_credential)?;
+    let expected_format = expected_signed_credential_format(credential_configuration)?;
+
+    if actual_format == expected_format {
+        Ok(())
+    } else {
+        Err(ApiError::builder(StatusCode::UNPROCESSABLE_ENTITY)
+            .title("Signed Credential Format Mismatch")
+            .type_url(type_url("issuance#signed-credential-format-mismatch"))
```
nanderstabel, line 307: See https://github.com/impierce/ssi-agent/pull/330/changes#r3472472755

agent_api_http/src/v0/issuance/credentials.rs
```diff
@@ -170,23 +243,139 @@ pub(crate) async fn credentials(
...
+fn expected_signed_credential_format(
+    credential_configuration: &CredentialConfigurationsSupportedObject,
+) -> Result<SignedCredentialFormat, ApiError> {
+    match &credential_configuration.credential_format {
+        CredentialFormats::JwtVcJson(_) => Ok(SignedCredentialFormat::JwtVcJson),
+        CredentialFormats::VcSdJwt(_) => Ok(SignedCredentialFormat::VcSdJwt),
+        CredentialFormats::DcSdJwt(_) => Ok(SignedCredentialFormat::DcSdJwt),
+        _ => Err(ApiError::builder(StatusCode::INTERNAL_SERVER_ERROR)
+            .title("Unsupported Credential Configuration Format")
+            .type_url(type_url("issuance#unsupported-credential-configuration-format"))
```
nanderstabel, line 327: See https://github.com/impierce/ssi-agent/pull/330/changes#r3472472755

agent_api_http/src/v0/issuance/credentials.rs
```diff
@@ -113,6 +170,11 @@ pub(crate) async fn credentials(
                 .finish());
         }
 
+        validate_signed_credential_format_matches_configuration(
+            credential.as_str().expect("signed credential must be a string"),
+            &credential_configuration,
+        )?;
```
nanderstabel: I'd suggest to rewrite this part to:
```rust
        let signed_credential = match credential.as_str() {
            Some(value) => value,
            None => {
                return Err(ApiError::builder(StatusCode::BAD_REQUEST)
                    .title("Invalid Credential Type")
                    .type_url(type_url("issuance#invalid-credential-type"))
                    .message("For signed credentials, the credential must be a string.")
                    .finish())
            }
        };

        validate_signed_credential_format_matches_configuration(signed_credential, &credential_configuration)?;
```

so there is no need to use `expect`

agent_api_http/src/v0/issuance/credentials.rs
```diff
@@ -170,23 +243,139 @@ pub(crate) async fn credentials(
...
+
+    serde_json::from_slice(&decoded).map_err(|_| {
+        invalid_signed_credential_format_error("Signed credential contains invalid JSON in its JWT segments.")
+    })
+}
+
+fn invalid_signed_credential_format_error(message: impl Into<String>) -> ApiError {
+    ApiError::builder(StatusCode::BAD_REQUEST)
+        .title("Invalid Signed Credential Format")
+        .type_url(type_url("issuance#invalid-signed-credential-format"))
```
nanderstabel, line 382: See https://github.com/impierce/ssi-agent/pull/330/changes#r3472472755

agent_api_http/src/v0/issuance/credentials.rs
```diff
@@ -256,24 +445,157 @@ pub async fn patch_credential(
...
+    explicit: &CredentialExpiry,
+    deadline: &CredentialExpiry,
+) -> Result<(), ApiError> {
+    match (explicit, deadline) {
+        // Template has no deadline — anything goes.
+        (_, CredentialExpiry::Never) => Ok(()),
+        // Explicit is "never" but template enforces a deadline.
+        (CredentialExpiry::Never, CredentialExpiry::Fixed(limit)) => Err(ApiError::builder(StatusCode::BAD_REQUEST)
+            .title("Expiration Exceeds Template Limit")
+            .type_url(type_url("issuance#expiration-exceeds-template-limit"))
```
nanderstabel, line 520: See https://github.com/impierce/ssi-agent/pull/330/changes#r3472472755

agent_api_http/src/v0/issuance/credentials.rs
```diff
@@ -256,24 +445,157 @@ pub async fn patch_credential(
...
+}
+
+/// Validates the credential data against the template's JSON Schema.///
+/// Returns a detailed error response if validation fails, listing all schema violations.
+fn validate_credential_against_schema(credential: &Value, schema: &Value) -> Result<(), Box<ApiError>> {
+    let validator = jsonschema::validator_for(schema).map_err(|e| {
+        Box::new(
+            ApiError::builder(StatusCode::INTERNAL_SERVER_ERROR)
+                .title("Invalid Template Schema")
+                .type_url(type_url("issuance#invalid-template-schema"))
```
nanderstabel, line 551: See https://github.com/impierce/ssi-agent/pull/330/changes#r3472472755

agent_api_http/src/v0/issuance/credentials.rs
```diff
@@ -256,24 +445,157 @@ pub async fn patch_credential(
...
+        .map(|e| format!("Path `{}`: {} (schema path: {})", e.instance_path(), e, e.schema_path()))
+        .collect();
+
+    if errors.is_empty() {
+        Ok(())
+    } else {
+        Err(Box::new(
+            ApiError::builder(StatusCode::UNPROCESSABLE_ENTITY)
+                .title("Credential Schema Validation Failed")
+                .type_url(type_url("issuance#credential-schema-validation-failed"))
```
nanderstabel, line 568: See https://github.com/impierce/ssi-agent/pull/330/changes#r3472472755

agent_api_http/src/v0/issuance/credentials.rs
```diff
@@ -303,30 +625,288 @@ pub mod tests {
...
+                "type": ["VerifiableCredential"],
+                "credentialSubject": {
+                    "id": "did:example:holder"
+                }
+            }
+        }));
+        let signature = URL_SAFE_NO_PAD.encode(b"signature");
+
+        format!("{header}.{payload}.{signature}")
+    }
```
nanderstabel, line 854: This function does not actually sign anything, so I think it should contain some comments explaining that this does not produce an actual (valid) signed credential

agent_api_http/src/v0/library/error.rs
nanderstabel: These errors still need Problem Detail documentation in `docs/problem-details/library.md`

agent_api_http/src/v0/templates/mod.rs
```diff
@@ -64,17 +63,17 @@ impl From<Template> for TemplateDto {
     }
 }
 
-#[derive(Deserialize, Serialize, Default, utoipa::ToSchema)]
-#[serde(default, rename_all = "camelCase")]
-pub struct CreateTemplateEndpointRequest {
-    pub title: Option<String>,
+#[derive(Debug, Deserialize, Serialize, Default, utoipa::ToSchema)]
+#[serde(default, deny_unknown_fields, rename_all = "camelCase")]
+pub struct CreateTemplateRequestBody {
```
nanderstabel: `CreateTemplateRequestBody` --> `CreateNewTemplateRequestBody`?

agent_api_http/src/v0/issuance/offers/mod.rs
```diff
@@ -39,57 +40,54 @@ pub(crate) async fn offers(
...
-                .title("No Credential Configuration Found")
-                .type_url(type_url("issuance#no-credential-configuration-found"))
-                .message(format!(
-                    "No Credential Configuration found with id: `{credential_configuration_id}`"
-                ))
-                .finish()
-        })?;
+    let authorization = Authorization {
+        pre_authorized: first_template.pre_authorized,
+        tx_code_constraints: None,
```
nanderstabel: Hardcoding `tx_code_constraints` to `None` is a regression (which means we will not be able to add pincode requirement to pre-authorized code flow offers). We can patch this back in later though

agent_api_http/src/v0/templates/mod.rs
```diff
@@ -59,13 +60,18 @@ impl From<Template> for TemplateDto {
...
             schema: *value.schema,
             schema_properties_attributes: value.schema_properties_attributes,
+            pre_authorized: value.pre_authorized,
         }
     }
 }
 
+fn default_pre_authorized() -> bool {
+    false
+}
```
nanderstabel: Are you sure default should be `false`? Previously in the `/v0/credential-configurations` the default was `true`, see:
```rust
#[derive(Deserialize, Serialize, Debug, Clone, PartialEq)]
pub struct Authorization {
    pub pre_authorized: bool,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub tx_code_constraints: Option<TxCodeConstraints>,
}

impl Default for Authorization {
    fn default() -> Self {
        Self {
            pre_authorized: true,
            tx_code_constraints: None,
        }
    }
}
```

If it is intended to make the default `false` for `pub pre_authorized: bool,`, then you do not need this function because default of `bool` is already `false`, so --> 
```rust
 #[serde(default)]
    pub pre_authorized: bool,
```

nanderstabel (dismissed):

nanderstabel (approved):

PR #340, opened 2026-05-07
feat(shared-kernel): extract request actor, add authorization hook

florishafkenscheid:
# Description of change
Added a shared-kernel authorization hook around application command and query execution.

This introduces neutral authorization primitives (`Actor`, `AuthorizationRequest`, `AuthorizationOperation`, and `AuthorizationChecker`) plus a default allow-all checker for existing open-source behavior. `ServiceHandle` can now dispatch commands and queries with optional actor context, and `ApplicationService` authorizes each command/query before calling the bounded context.

The goal is to make authorization enforceable below inbound adapters such as HTTP, GraphQL, gRPC, or queues, rather than tying access checks to a specific API transport.

N/A

Verified locally with:

```sh
cargo fmt --check
cargo test -p shared-kernel
cargo clippy -p shared-kernel -- -D warnings
cargo check --workspace
git diff --check
```

The shared-kernel tests cover default allow-all behavior, actor context propagation, authorization request construction, denied command/query execution, error mapping, dropped reply channels, and service shutdown.


agent_api_http/src/v0/holder/holder/credentials/mod.rs
```diff
@@ -80,6 +89,7 @@ pub(crate) async fn post_credentials(
 #[axum_macros::debug_handler]
 pub(crate) async fn credential(
     State(state): State<Arc<HolderState>>,
+    _actor: Option<Extension<Option<Actor>>>,
```
nanderstabel: For those endpoints where `Actor` is not being used, let's just omit this line 👌

agent_api_http/src/v0/openapi.rs
```diff
@@ -33,6 +33,7 @@ mod tests {
 
     /// Generates an openapi.yaml file from the annotations in the code.
     #[test]
+    #[ignore = "run manually to update the committed OpenAPI specification"]
```
nanderstabel: I don't think it is a good idea to ignore this test because we have a step in the `format-lint-test.yaml` workflow that checks whether any generated files should be updated (including the OpenAPI file):
```yaml
      - name: Generated files are unchanged
        run: git diff --exit-code
```

agent_api_http/src/error.rs
```diff
@@ -88,6 +95,22 @@ where
...
 
+impl<T: IntoApiErrorExt> IntoApiErrorExt for CommandHandlerError<T>
+where
+    T: Send + Sync + 'static,
+{
+    fn into_api_error(self) -> ApiError {
+        match self {
+            CommandHandlerError::Forbidden => ApiError::builder(StatusCode::FORBIDDEN)
+                .title("Forbidden")
+                .type_url(type_url("authorization#forbidden"))
```
nanderstabel: This will add a URL to the [Problem Details Error Response](https://www.rfc-editor.org/rfc/rfc9457.html#name-type). The purpose for this is to describe more details about the error and how it can be resolved. The Problem Details that we have described so far can be found here: https://github.com/impierce/ssi-agent/tree/beta/docs/problem-details

You can add authorization.md file there and add the Problem Detail description for this error response :)

agent_api_http/src/v0/templates/mod.rs
```diff
@@ -503,6 +607,220 @@ pub(crate) async fn delete_template(
...
+
+    fn post_json(uri: &str, body: serde_json::Value) -> Request<Body> {
+        Request::builder()
+            .method("POST")
+            .uri(uri)
+            .header(header::CONTENT_TYPE, "application/json")
+            .body(Body::from(body.to_string()))
+            .unwrap()
+    }
+}
```
nanderstabel, line 1056: Do these tests specifically test Templates, or do they test Authorization? If Templates, then I think adding these tests is out of scope for this PR, if Authorization, then I think you have already tests elsewhere that cover Authorization behavior, and it is not necessary to test that again here.

agent_api_http/src/v0/issuance/credential_issuer/credential.rs
```diff
@@ -37,6 +39,7 @@ const POLLING_INTERVAL_MS: u64 = 100;
 #[axum_macros::debug_handler]
 pub(crate) async fn credential(
     State(state): State<Arc<IssuanceState>>,
+    actor: Option<Extension<Option<Actor>>>,
```
nanderstabel: This endpoint is a Public Endpoint which is part of the OpenID4VCI flow, which means that it should not require any authorization outside of what is defined in the standard. The rule is basically that the authorization check that you have build should only be applied to the Management API which is basically every endpoint behind the `/v0` path segment :) For example in `issuance::router`:
```rust
pub fn router(issuance_state: Arc<IssuanceState>) -> Router {
    Router::new()
        .nest(
            // All endpoints nested behind `API_VERSION` must include the actor Extraction
            API_VERSION,
            Router::new()
                .route("/credentials", post(credentials).get(all_credentials))
                .route(
                    "/credentials/{credential_id}",
                    get(credentials::credential).patch(patch_credential),
                )
                .route("/credential-configurations", post(credential_configurations))
                .route("/offers", post(offers).get(all_offers))
                .route("/offers/{offer_id}", get(offer))
                .route("/offers/send-offer-to-individual", post(individual_offer))
                .route("/offers/send-offer-to-organization", post(organization_offer)),
        )
        // All other endpoints must not include the actor Extraction
        .route(
            "/.well-known/oauth-authorization-server",
            get(oauth_authorization_server),
        )
        .route("/.well-known/openid-credential-issuer", get(openid_credential_issuer))
        .route("/openid4vci/credential", post(credential))
        .route("/openid4vci/nonce", post(nonce))
        .route("/openid4vci/notification", post(notification))
        .route("/openid4vci/credential-offer/{offer_id}", get(credential_offer_uri))
        .route("/ietf-oauth-token-status-list/{path}", get(token_status_list))
        // TODO: Move this route to `../library` once `agent_library` is properly implemented.
        .route(
            "/vct/{credential_configuration_id}/{version}",
            get(ietf_oauth_sd_jwt_vc::type_metadata),
        )
        .with_state(issuance_state)
}
```

agent_identity/src/state.rs
```diff
@@ -579,3 +700,127 @@ pub async fn query_all_documents(
...
+                all_documents: Arc::new(AllDocumentsRepo),
+                profile: none_repo(),
+                service: none_repo(),
+                all_services: none_repo(),
+            },
+        };
+
+        publish_decentrally_hosted_documents(&state).await.unwrap();
+    }
+}
```
nanderstabel: I have a feeling these tests are out of scope for this PR, correct?

agent_shared/src/handlers.rs
```diff
@@ -29,25 +30,249 @@ where
...
+) -> Result<(), CommandHandlerError<<A as Aggregate>::Error>>
 where
     A: Aggregate,
     <A as Aggregate>::Command: Send + Sync + std::fmt::Debug,
 {
+    let authorization_request = AuthorizationRequest {
+        actor,
+        operation: AuthorizationOperation::Command {
+            aggregate_id: aggregate_id.to_string(),
+            command_type: std::any::type_name::<A::Command>(),
```
nanderstabel, line 97: `std::any::type_name::<A::Command>()` will create a string containing the name of the Command enum, and not of specific Variants of the Command, so that is not very granular. I could imagine that an Actor has permission to execute Command variant A, B, C, but not D, E, and F for example. 

Could also be a future improvement (@daniel-mader ?), if so, then I think it would be nice to mention that in a `TODO` comment here.

agent_store/Cargo.toml
```diff
@@ -18,9 +18,14 @@ cqrs-es.workspace = true
...
 sqlx = { version = "0.8", features = [
     "postgres",
     "runtime-tokio-rustls",
     "json",
 ] }
 tokio.workspace = true
+
+[dev-dependencies]
+agent_secret_manager = { path = "../agent_secret_manager", features = ["test_utils"] }
+agent_shared = { path = "../agent_shared", features = ["test_utils"] }
```
nanderstabel: These dependencies don't seem to be used anywhere in `agent_store` so they can be omitted here :)

shared-kernel/src/application_service.rs
```diff
@@ -152,10 +205,29 @@ async fn process_command<AC: ApplicationContext>(context: &AC, msg: CommandEnvel
...
+) {
     debug!("Processing query");
 
-    let result = context.handle_query(msg.query).await;
+    let authorization_request = AuthorizationRequest {
+        actor: msg.actor.clone(),
+        operation: AuthorizationOperation::Query {
+            query_type: std::any::type_name::<AC::Query>(),
+        },
+    };
```
nanderstabel, line 230: I see here that Authorization is not only checked for Commands, but for Queries as well makes a lot of sense I think. However, this is then currently inconsistent with query endpoints in the OS part since `fn query_handler` does not contain the `authorization_checker` like `fn command_handler` does now. 

If you indeed update `fn query_handler` the same way as `fn command_handler`, then my [other comment](https://github.com/impierce/ssi-agent/pull/340/changes#r3286464631) does not apply anymore since you can then actually supply `actor` to `fn query_handler`.

shared-kernel/src/authorization.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+pub struct AllowAllAuthorizationChecker;
+
+#[async_trait]
+impl AuthorizationChecker for AllowAllAuthorizationChecker {
+    async fn is_authorized(&self, _request: &AuthorizationRequest) -> bool {
+        true
+    }
+}
```
nanderstabel, line 116: These types and traits are quite important, but also a bit hard to understand without context so I think it would be good to add some doc comments.

shared-kernel/src/authorization.rs
```diff
@@ -0,0 +1,58 @@
...
+impl ActorExtractor for NoActorExtractor {
+    fn extract_actor(&self, _input: &dyn ToActor) -> Option<Actor> {
+        None
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AuthorizationOperation {
+    Command {
+        aggregate_id: String,
```
nanderstabel, line 71: Just out of curiosity, what is the purpose of also adding the `aggregate_id` here? Is it for example to make it possible to give permissions per aggregate 'instance'? What I mean by the goal to make it possible for an Admin to declare that Actor A is allowed to perform commands on Credential A, but not Credential B?

nanderstabel (changes requested):

nanderstabel (changes requested):
Some more places in the code need to utilize a public command handler, see: https://github.com/impierce/ssi-agent/commit/20734ae85245dd90ea3b61caf5cbcb2421ddb65c

shared-kernel/src/authorization.rs
```diff
@@ -19,17 +19,19 @@ pub trait ToActor {
 }
 
 /// Extracts an [`Actor`] from an input object.
+#[async_trait]
 pub trait ActorExtractor: Send + Sync + 'static {
     /// Returns the actor that should be attached to the application operation.
-    fn extract_actor(&self, input: &dyn ToActor) -> Option<Actor>;
+    async fn extract_actor(&self, input: &(dyn ToActor + Sync)) -> Option<Actor>;
```
nanderstabel: nit: You've already added `Sync` as a supertrait of `ToActor` so you do not need to add it again in these method signatures. So you can simply change all occurrences of `&(dyn ToActor + Sync)` back to `dyn ToActor`

nanderstabel (dismissed):

daniel-mader (dismissed):

daniel-mader (approved):

Issue #341, opened 2026-05-08
[Bug] InvalidProof error wherre I expected an InavlidToken


berkes:
### Description

https://github.com/impierce/ssi-agent/commit/cb9439c7481518e3dd3302ebf39c17666005d78d#diff-4bc405d13deedc25d81e3758ef4c16f442ecd5fed7fd1d4772665f13e3c2213dL42-L299 in commit cb9439c7481518e3dd3302ebf39c17666005d78d changed the error returned when access token is wrong, from InvalidToken to InvalidProof.

This then comes back as a 400, where I would expect a 403.

Aside: when I look at the mapping of the error to the HTTP errors, I see ` InvalidProofError(_) => ApiError::new(StatusCode::INTERNAL_SERVER_ERROR)` so purely from the code I'd expect a 500, probably I am misunderstanding that?

Anyway, could be that this change was intentional. Though it was changed in a commit that `bump OID4VCI dependencies ` , so I have a gut-feeling it was accidental?

It's not a big deal, but it did cause me some confusion and digging, because I thought my proof was wrong, where it really was the authentication that was wrong. 

### Hardware Specification

Operating system:
RAM: 
Cores:
Device: 

### Steps to Reproduce the Bug

1. Make an offer
2. Walk through the authentication code flow, stop just before requesting the credential
3. Request the credential with a valid proof, but an invalid access token; could be just any string really.

### Expected Behaviour

I get a 403, with `{ "error": "invalid_token" }` 

### Actual Behaviour

I get a 400, with `{ "error": "invalid_proof" }` 

### Errors

-

nanderstabel:
Thank you for reporting @berkes , @coplat is looking in to this :)

coplat:
Hey @berkes, thanks for reporting. A fix is in review now. You're right that it was accidental. According to the spec, access_token errors should return an authorization error response (as you rightly expected, an `"invalid_token"`) as defined in [[RFC6750](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-1_0-wg-draft.html#RFC6750)]. Small note, RFC 6750 actually specifies HTTP 401 (Unauthorized) as opposed to 403 for this case, so that's what the fix returns.

PR #343, opened 2026-05-12
fix: access token and credential endpoint error responses

coplat:
# Description of change
Previously, access token validation failures on the credential endpoint returned a 400 with `invalid_proof`, which was incorrect, as it mixed up authorization errors (defined in [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1), section 3) with credential request errors ([OID4VCI 1.0 section 8.3.1.2](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-1_0-wg-draft.html#name-credential-request-errors)). A similar mistake was also caught in the notification endpoint. 

Access token validation failures now return 401 with` {"error": "invalid_token"}` and a WWW-Authenticate header per [RFC 6750 Section 3.](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1)

Missing issuer_state now returns 400 with `{"error": "invalid_credential_request"}` per OID4VCI Section 8.3.1.2

Updated the IntoResponse impl for PublicError::AccessTokenError from 500 to the correct 401 with RFC 6750 response format. 

This PR also corrects the internal API error mappings for the credential endpoint errors (e.g. MissingProofError, InvalidProofError) from `INTERNAL_SERVER_ERROR` to `BAD_REQUEST`, and expanded the public error mappings to return proper OID4VCI error responses. 

fixes issue #341 

### DIFF
--- a/agent_api_http/src/v0/issuance/credential_issuer/credential.rs
+++ b/agent_api_http/src/v0/issuance/credential_issuer/credential.rs
@@ -40,12 +40,12 @@ pub(crate) async fn credential(
     AuthBearer(access_token): AuthBearer,
     Json(credential_request): Json<CredentialRequest>,
 ) -> Result<Response, PublicError> {
-    let offer_id = AccessTokenValidationService::validate(&state, &access_token)
-        .await
-        .ok()
-        // The Access Token must contain the `issuer_state` claim, which is used to identify the `offer_id`.
-        .and_then(|claims| claims.issuer_state)
-        .ok_or_else(|| PublicError::from(CredentialErrorResponse::InvalidProof))?;
+    let claims = AccessTokenValidationService::validate(&state, &access_token).await?;
+    // The Access Token must contain the `issuer_state` claim, which is used to identify the `offer_id`.
+
+    let offer_id = claims
+        .issuer_state
+        .ok_or_else(|| PublicError::from(CredentialErrorResponse::InvalidCredentialRequest))?;
 
     NonceValidationService::validate(&state, &credential_request)
         .await
--- a/agent_api_http/src/v0/issuance/credential_issuer/credential.rs
+++ b/agent_api_http/src/v0/issuance/credential_issuer/credential.rs
@@ -7,7 +7,8 @@ use crate::{
 };
 use agent_issuance::{
     application::{
-        access_token_validation_service::AccessTokenValidationService, nonce_validation_service::NonceValidationService,
+        access_token_validation_service::{AccessTokenValidationError, AccessTokenValidationService},
+        nonce_validation_service::NonceValidationService,
     },
     credential::{command::CredentialCommand, views::CredentialView},
     offer::{command::OfferCommand, views::OfferView},
@@ -41,11 +42,11 @@ pub(crate) async fn credential(
     Json(credential_request): Json<CredentialRequest>,
 ) -> Result<Response, PublicError> {
     let claims = AccessTokenValidationService::validate(&state, &access_token).await?;
-    // The Access Token must contain the `issuer_state` claim, which is used to identify the `offer_id`.
 
+    // The Access Token must contain the `issuer_state` claim, which is used to identify the `offer_id`.
     let offer_id = claims
         .issuer_state
-        .ok_or_else(|| PublicError::from(CredentialErrorResponse::InvalidCredentialRequest))?;
+        .ok_or_else(|| PublicError::from(AccessTokenValidationError::InvalidToken))?;
 
     NonceValidationService::validate(&state, &credential_request)
         .await
--- a/agent_api_http/src/v0/issuance/credential_issuer/notification.rs
+++ b/agent_api_http/src/v0/issuance/credential_issuer/notification.rs
@@ -28,9 +28,7 @@ pub async fn notification(
     let notification_request: NotificationRequest = serde_json::from_value::<NotificationRequest>(raw_value)
         .map_err(|_| PublicError::from(NotificationErrorResponse::InvalidNotificationRequest))?;
 
-    let _claims = AccessTokenValidationService::validate(&state, &access_token)
-        .await
-        .map_err(|_err| PublicError::from(NotificationErrorResponse::InvalidNotificationRequest))?;
+    let _claims = AccessTokenValidationService::validate(&state, &access_token).await?;
 
     let credentials = match query_handler("all_credentials", &state.query.all_credentials).await? {
         Some(all_credentials) => all_credentials.credentials,
@@ -145,7 +143,7 @@ mod tests {
             name: &'static str,
             access_token: String,
             payload: String,
-            expected_error: NotificationErrorResponse,
+            expected_status: StatusCode,
         }
 
         let test_cases = vec![
@@ -158,7 +156,7 @@ mod tests {
                     event_description: None,
                 })
                 .unwrap(),
-                expected_error: NotificationErrorResponse::InvalidNotificationId,
+                expected_status: NotificationErrorResponse::InvalidNotificationId.status_code(),
             },
             TestCase {
                 name: "Invalid Access Token",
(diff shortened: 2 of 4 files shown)


agent_api_http/src/v0/issuance/credential_issuer/credential.rs
```diff
@@ -40,12 +40,12 @@ pub(crate) async fn credential(
...
-        .ok()
-        // The Access Token must contain the `issuer_state` claim, which is used to identify the `offer_id`.
-        .and_then(|claims| claims.issuer_state)
-        .ok_or_else(|| PublicError::from(CredentialErrorResponse::InvalidProof))?;
+    let claims = AccessTokenValidationService::validate(&state, &access_token).await?;
+    // The Access Token must contain the `issuer_state` claim, which is used to identify the `offer_id`.
+
+    let offer_id = claims
+        .issuer_state
+        .ok_or_else(|| PublicError::from(CredentialErrorResponse::InvalidCredentialRequest))?;
```
nanderstabel: The errors defined [here](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-1_0-wg-draft.html#section-8.3.1.2-3.1.1) only apply to errors related to the Request's payload (request body):
> For errors related to the Credential Request's payload [...] the specific error codes from this section MUST be used instead of the generic invalid_request parameter defined in Section 3.1 of RFC6750

but the `issuer_state` is part of the Bearer token (== `access_token`) which is a Request Header (so it is not part of the request body). Therefore I think here we need to fall back to section 3.1 in rfc6750 and then I think the error below then is the most applicable, wyt (malformed since for us the access_token JWT MUST include the `issuer_state` parameter)?
>   invalid_token
         The access token provided is expired, revoked, **malformed, or
         invalid for other reasons**.  The resource SHOULD respond with
         the HTTP 401 (Unauthorized) status code.  The client MAY
         request a new access token and retry the protected resource
         request.

agent_api_http/src/v0/issuance/credential_issuer/notification.rs
```diff
@@ -28,9 +28,7 @@ pub async fn notification(
     let notification_request: NotificationRequest = serde_json::from_value::<NotificationRequest>(raw_value)
         .map_err(|_| PublicError::from(NotificationErrorResponse::InvalidNotificationRequest))?;
 
-    let _claims = AccessTokenValidationService::validate(&state, &access_token)
-        .await
-        .map_err(|_err| PublicError::from(NotificationErrorResponse::InvalidNotificationRequest))?;
+    let _claims = AccessTokenValidationService::validate(&state, &access_token).await?;
```
nanderstabel: See https://github.com/impierce/ssi-agent/pull/343/changes#r3232349771

agent_api_http/src/v0/issuance/error.rs
```diff
@@ -102,13 +102,13 @@ impl IntoApiErrorExt for OfferError {
...
-            MissingCredentialConfigurationIdsError => ApiError::new(StatusCode::INTERNAL_SERVER_ERROR),
-            UnknownCredentialConfiguration(_) => ApiError::new(StatusCode::INTERNAL_SERVER_ERROR),
-            UnsupportedCredentialIdentifierError => ApiError::new(StatusCode::INTERNAL_SERVER_ERROR),
+            MissingCredentialError => ApiError::new(StatusCode::BAD_REQUEST),
+            MissingProofError => ApiError::new(StatusCode::BAD_REQUEST),
+            InvalidProofError(_) => ApiError::new(StatusCode::BAD_REQUEST),
+            MissingProofIssuerError => ApiError::new(StatusCode::BAD_REQUEST),
+            MissingCredentialConfigurationIdsError => ApiError::new(StatusCode::BAD_REQUEST),
+            UnknownCredentialConfiguration(_) => ApiError::new(StatusCode::BAD_REQUEST),
+            UnsupportedCredentialIdentifierError => ApiError::new(StatusCode::BAD_REQUEST),
```
nanderstabel: These type of errors are specific to the Public API endpoints (`/openid4vci/credential` and `/openid4vci/notification`) so they will logically never occur in the management API (everything behind `/0`):
- Management API (UniCore specific) --> ApiError (Problem Details)
- Public API (standard/specification specific --> PublicError (as defined in the corresponding spec)

IF these errors somehow do occur in one of the Management API endpoints, this is a server error because it logically should never be able to happen, ergo it should return `INTERNAL_SERVER_ERROR`. We _could_ also map those error variants to `=> unreachable!()`, however, if the code somehow does still end up here the server would crash, which is worse than returning `INTERNAL_SERVER_ERROR`. Returning `BAD_REQUEST` to clients here signals that whatever goes wrong is due to the clients input, which is misleading. 

Example for `MissingProofError`:
The `proof` as a concept is only relevant in the context of the issuance flow in OpenID4VCI (== Standard/Specification), so it can only be 'missing' in that context. When calling the Issuance Management API endpoints (like `/v0/credentials`, `/v0/offers`) this error can logically never occur, because `proof` is not a thing in the Management API context. Still, due to Rust's type-safety, we need to map all `CredentialError` to both `PublicError` and `ApiError`, simply because a subset `CredentialError`'s variants relevant is to Management API and another subset relevant is to Public API.

PS: this all IS indeed confusing so when we have the time we should design a better error handling flow.

agent_api_http/src/v0/issuance/credential_issuer/credential.rs
nanderstabel: So I guess we need something like this?:
```rust
    let offer_id = claims
        .issuer_state
        .ok_or_else(|| PublicError::from(AccessTokenValidationError::MissingIssuerState))?;
```

I guess it's a bit weird though to invoke an `AccessTokenValidationError` error outside of the context of `AccessTokenValidationService` but whatever, I think it is acceptable.

agent_api_http/src/v0/issuance/error.rs
```diff
@@ -212,16 +217,39 @@ impl IntoPublicError for OfferError {
...
+            }
+            UnknownCredentialConfiguration(_) => PublicError::CredentialError(OID4VCError::new(
+                CredentialErrorResponse::UnknownCredentialConfiguration,
+            )),
+            UnsupportedCredentialIdentifierError => {
+                PublicError::CredentialError(OID4VCError::new(CredentialErrorResponse::UnknownCredentialIdentifier))
+            }
+
+            // Internal errors that shouldn't reach the public API
             _ => PublicError::InternalServerError,
```
nanderstabel: Using the `_ =>` fallback is dangerous/misleading because it makes it easy to forget to map Domain Errors to `PublicError`/`ApiError` when you add new variants (which I assume is why you are adding the mappings above. It is better to not use the `_` fallback and map every error variant explicitly so that the Rust Analyzer will tell you to add a mapping here as well when a new Domain Error variant is added. So:
```rust
            // Internal errors that shouldn't reach the public API
            SendCredentialOfferError(_) => PublicError::InternalServerError,
            UnsupportedTokenRequestGrantTypeError => PublicError::InternalServerError,
            InvalidCredentialOfferUriError(_) => PublicError::InternalServerError,
```

nanderstabel (changes requested):

agent_api_http/src/v0/issuance/credential_issuer/credential.rs
coplat: Hm this makes sense. W could also reuse InvalidToken here since a missing issuer_state falls under 'malformed', so it keeps it close to the RFC 6750 wording. Or would you suggest the more specific "MissingIssuerState?"

agent_api_http/src/v0/issuance/credential_issuer/notification.rs
coplat: Here the accesstokenvalidationservice already correctly returns invalid_token.

agent_api_http/src/v0/issuance/error.rs
coplat: Yeah it is a little bit confusing and I recall now having an issue with this before. Thanks for the explanation!

agent_issuance/src/application/access_token_validation_service.rs
```diff
@@ -31,6 +31,7 @@ pub enum AccessTokenValidationError {
     InvalidToken,
     #[error("Failed to resolve the public key for the token's `kid`")]
     KidResolutionError,
+
```
nanderstabel: `cargo fmt` is complaining about this empty line. Once this is resolved I think this PR is good to go :)

nanderstabel (changes requested):

agent_api_http/src/v0/issuance/credential_issuer/credential.rs
nanderstabel: Yeah simply reusing InvalidToken makes sense :ok_hand:

nanderstabel (approved):

PR #347, opened 2026-05-26
feat: introduce catalogs

coplat:
# Description of change
Introduces a catalogs module in `agent_library`. This new feature allows the grouping of templates within catalogs, and initializes the following corresponding endpoints:

`POST` `/create-new-catalog` -> create a catalog given CatalogDisplay properties (name, description and logo).
`POST` `/add-templates-to-catalog` -> Add one or more templates to an existing catalog.
`POST` `/remove-templates` -> Remove one or more templates from an existing catalog.
`POST` `/change-catalog-appearance` -> Modify the name, description or icon of a catalog.
`POST` `/make-catalog-public` -> Change the catalogs visibility to public.
`POST` `/make-catalog-private` -> Change the catalogs visibility to private.


agent_api_http/src/v0/library/catalog/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+use http::StatusCode;
+use http_api_problem::ApiError;
+use hyper::header;
+use serde::{Deserialize, Serialize};
+use std::sync::Arc;
+
+/// Data transfer object for Catalogs.
+#[derive(Serialize, Deserialize, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase")]
+#[schema(as = Catalog)]
```
daniel-mader, line 29: This is redundant if the struct has the same name. Did you forget to rename the struct to `CatalogDto`?

agent_api_http/src/v0/library/catalog/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+pub struct Catalog {
+    #[serde(rename = "id")]
+    pub catalog_id: String,
+    pub display: CatalogDisplay,
+    pub template_ids: Vec<String>,
+    pub visibility: CatalogVisibility,
+    pub modified_at: DateTime<Utc>,
+    pub is_deleted: bool,
```
daniel-mader: A DTO shouldn't hold the `is_deleted` flag since it's exposing an inner state in the OpenAPI.

agent_api_http/src/v0/library/catalog/openapi.rs
```diff
@@ -0,0 +1,20 @@
...
+};
+
+use utoipa::OpenApi;
+
+#[derive(OpenApi)]
+#[openapi(
+    paths(create_catalog, delete_catalog, make_catalog_public, make_catalog_private, add_templates, remove_templates, update_display, get_all_catalogs, get_catalog_by_id),
+    tags(
+        (name = "Library", description = "Manage your own templates, browse and import external templates."),
+        (name = "Catalogs", description = "Create and manage catalogs to organize and share your templates.")
```
daniel-mader: You're using the tag `tags = ["Library", "Catalog"]` for all endpoints, but here `"Catalogs"` is documented. Please update to `"Catalog"` as well.

agent_api_http/src/v0/library/catalog/openapi.rs
```diff
@@ -0,0 +1,20 @@
...
+    get_all_catalogs::__path_get_all_catalogs, get_catalog_by_id::__path_get_catalog_by_id,
+};
+
+use utoipa::OpenApi;
+
+#[derive(OpenApi)]
+#[openapi(
+    paths(create_catalog, delete_catalog, make_catalog_public, make_catalog_private, add_templates, remove_templates, update_display, get_all_catalogs, get_catalog_by_id),
+    tags(
+        (name = "Library", description = "Manage your own templates, browse and import external templates."),
```
daniel-mader: `"Library"` tag is not needed here since it's already defined in `templates/openapi.rs`.

agent_api_http/src/v0/library/error.rs
```diff
@@ -114,6 +115,28 @@ impl IntoApiErrorExt for TemplateError {
     }
 }
 
+impl IntoApiErrorExt for CatalogError {
+    fn into_api_error(self) -> ApiError {
+        match self {
+            CatalogError::TemplatesNotFound(_) => ApiError::builder(StatusCode::NOT_FOUND)
```
daniel-mader: This is called TemplatesNotFound, but the error describes "Template Not Found". Is it thrown when a single template is not found?

agent_api_http/src/v0/library/error.rs
```diff
@@ -114,6 +115,28 @@ impl IntoApiErrorExt for TemplateError {
...
+                .title("Template Not Found")
```
daniel-mader, line 122: Those errors are all missing a `type_url` and a `message` and a documentation in `docs/problem-details`.

agent_api_http/src/v0/library/error.rs
```diff
@@ -114,6 +115,28 @@ impl IntoApiErrorExt for TemplateError {
...
+                .source(self)
+                .finish(),
+            CatalogError::MissingField(_) => ApiError::builder(StatusCode::BAD_REQUEST)
```
daniel-mader: I think this is too generic. The only information a caller would get right now is:

```
400 Bad Request

Missing Required Field
```

agent_api_http/src/v0/library/error.rs
```diff
@@ -114,6 +115,28 @@ impl IntoApiErrorExt for TemplateError {
...
+                .title("Missing Required Field")
+                .source(self)
+                .finish(),
+            CatalogError::CatalogNotFound(_) => ApiError::builder(StatusCode::NOT_FOUND)
+                .title("Catalog Not Found")
+                .source(self)
+                .finish(),
+            CatalogError::DuplicateTemplate(_) => ApiError::builder(StatusCode::BAD_REQUEST)
```
daniel-mader: Should this really be an error? You could just make a unique set out of the template_ids passed and it would use only the unique ones. We can also leave it like that, but I think it's overly strict.

agent_library/src/catalog/views/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+#[derive(Debug, Clone, Deserialize, Default, Serialize, ToSchema)]
+#[schema(as = Catalog)]
+pub struct CatalogView {
+    #[serde(rename = "id")]
+    pub catalog_id: String,
+    pub display: CatalogDisplay,
+    pub template_ids: Vec<String>,
+    pub visibility: CatalogVisibility,
+    pub modified_at: DateTime<Utc>,
+    pub deleted: bool,
```
daniel-mader, line 18: This should not be returned in the view. A `deleted` flag is an internal flag to track state. This view should always be updated when a catalog is removed, so the flag would be redundant anyway. If you use this view to track state, then `#[serde(skip)]` it (less clean).

agent_library/src/catalog/services.rs
```diff
@@ -0,0 +1,32 @@
...
+    async fn template_exists(&self, id: &str) -> bool;
+    async fn missing_templates(&self, ids: &[String]) -> Vec<String>;
+}
+
+pub struct CatalogServiceImpl {
+    pub template_view_repo: Arc<dyn ViewRepository<TemplateView, Template>>,
+}
+
+#[async_trait]
+impl CatalogServices for CatalogServiceImpl {
```
daniel-mader, line 17: I don't understand this part of the code. Why are there two functions? Is the `template_exists` only a helper for tests? If yes, it shouldn't be exposed a regular service function.

agent_library/src/catalog/services.rs
```diff
@@ -0,0 +1,32 @@
...
+    async fn template_exists(&self, id: &str) -> bool {
+        self.template_view_repo.load(id).await.ok().flatten().is_some()
+    }
+
+    async fn missing_templates(&self, ids: &[String]) -> Vec<String> {
```
daniel-mader: I couldn't tell from the name of the function what it does? The only thing it does is looping over the ids and calling the other function, right? Wouldn't `check_all_templates_exist()` would be more speaking then?

agent_api_http/openapi-generated.yaml
```diff
@@ -9,6 +9,48 @@ servers:
...
+              schema:
+                $ref: '#/components/schemas/Catalog'
+  /v0/change-catalog-appearance:
+    post:
+      tags:
+      - Library
+      - Catalog
+      summary: Changes a catalog's display information
+      description: Changes a catalog's display information such as name, description, and icon.
+      operationId: update_display
```
daniel-mader: I think the entire OpenAPI needs another refinement to align the endpoint names, operationIds and command names inside Rust. Right now we have:

```
/v0/change-catalog-appearance

operationId: update_display

CatalogCommand::UpdateDisplay
```

This applies to all endpoint definitions.

agent_api_http/openapi-generated.yaml
```diff
@@ -247,6 +310,21 @@ paths:
...
+      summary: Delete a catalog
+      description: Deletes a catalog.
+      operationId: delete_catalog
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/DeleteCatalogRequest'
+        required: true
+      responses: {}
```
daniel-mader: This produces an invalid OpenAPI, since responses can not be empty.

agent_library/src/catalog/command.rs
```diff
@@ -0,0 +1,33 @@
+use crate::catalog::aggregate::{CatalogDisplay, CatalogVisibility};
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize)]
+#[serde(untagged)]
+pub enum CatalogCommand {
+    CreateCatalog {
+        catalog_id: String,
```
daniel-mader, line 8: The `catalog_id` should be generated by UniCore and not be allowed to be passed in when a new catalog is created.

docs/problem-details/library.md
```diff
@@ -73,3 +75,17 @@ This error occurs when `schemaPropertiesAttributes` contains duplicate keys afte
 ## Draft Template Cannot Be Public
 
 This error occurs when a template in `Draft` stage is made public.
+
+# Catalogs 
+
+## Template Not Found
```
daniel-mader: Have you checked if `.type_url(type_url("library#template-not-found"))` links still work, if you introduce headlines, such as `# Catalogs`?

agent_library/src/catalog/command.rs
coplat: This is currently not possible in ssi-agent as the command_handler requires an aggregate_id as its first argument.

agent_library/src/catalog/views/mod.rs
coplat: Since the deleted flag is not returned in the CatalogDTO we don't need to serde(skip) here.

docs/problem-details/library.md
coplat: This still works. Have renamed it to catalog-template-not-found just to differentiate between the two

daniel-mader (approved):

